PR #19, opened 2011-06-30
check to see if file exists before attempting require_once in Loader

jdodds:
this skips over files that are obviously not meant to be required,
such as temp files from editors when a file is open in a datafixtures
directory with unsaved changes in it, while preserving the "shut
down everything" behaviour of "require".

I figure that script-halting behavior could actually be beneficial here,
and that that's why you're not using include_once. If the behavior of
file_exists wrt safe_mode is a problem, there are other ways of checking.


lavoiesl:
Closing as obsolete

PR #44, opened 2012-03-09
Add ability to exclude tables in purge

saem:
At work we manage our database change sets through [Liquibase](http://liquibase.org/) (this isn't negotiable). We require fixtures to load test data. The database is initialized (both structure and seed data) via Liquibase.

After which we'd like to load fixture data for running integration tests with, having consistent test data for all of our developers, and standardized data for manual regression testing. The problem is if we have any pre-seeded dictionaries in the database (we have a few), which are referenced by our entities, they get deleted/truncated (well not truncated since it doesn't work in newer versions of mysql). Rather than duplicating the fixtures into doctrine from liquibase, or some other hack. It'd be ideal if we could simply provide the purger with a suppression/exclusion list.

I've added basic purger tests, and an initial implementation and test case for this.


lavoiesl:
Do we still want this? If so, it needs to be rebased

totaltrash:
I think this can be closed - it was done in #225

PR #72, opened 2012-07-26
Update lib/Doctrine/Common/DataFixtures/Executor/MongoDBExecutor.php

weyandch:
When loading fixtures with MongoDB the indexes are getting dropped on purging the collections.

Added the proper command to update the Indexes right after the purging, to ensure consistency


lavoiesl:
Closed in favour of #118

PR #74, opened 2012-09-13
New ProvideFixtureInterface 

djlambert:
I was thinking it would be much easier to define fixture dependencies than manually setting an order. After looking at the code I found there was an existing DependentFixtureInterface that was undocumented. The new ProvideFixtureInterface allows the use of a simple string instead of the full fixture class name in the array returned from getDependencies. Documentation is also included.

The test ProxyReferenceRepositoryTest::testReferenceReconstruction is failing, but it's also failing on master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ or define alias names with ProvideFixtureInterface:
 
         public function getDepends()
         {
-            return array('Other'); // fixture class(es) or provide string(s) fixture is dependent on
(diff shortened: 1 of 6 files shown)


lavoiesl:
testReferenceReconstruction error will be fixed by #177

lavoiesl:
It seems that this PR is obsolete since we now use DependentFixtureInterface, can we close this?

PR #79, opened 2012-10-13
Method for loading fixtures from file added.

mariusz-alef-bak:
This PR allows to load fixtures from a single file by adding `loadFromFile` method to `Loader` class.

In some projects which require various fixture sets (e.g. for different setups for various environments) being able to point fixtures only by directory is quite a pain as some fixtures may be reused in few sets. Moving them to subdirectories doesn't solve the problem in some cases as namespaces would have to be adjusted accordingly. There were some issues reported concerning the lack of this feature both in this repo and in DoctrineFixturesBundle. They didn't however contain any code needed to accomplish the goal.

I am planning also to provide a related PR to DoctrineFixturesBundle when this one meets your acceptance.

PHPUnit tests fail in one case, but it does not seem related and it fails also on `master`. Namely:
```
There was 1 failure:

1) Doctrine\Tests\Common\DataFixtures\ProxyReferenceRepositoryTest::testReferenceReconstruction
Failed asserting that Doctrine\Tests\Common\DataFixtures\TestEntity\Role Object (...) is an instance of class "Doctrine\ORM\Proxy\Proxy".

/home/mariusz/Development/data-fixtures/tests/Doctrine/Tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php:116
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ You can load a set of fixtures from a directory as well:
 
     $loader->loadFromDirectory('/path/to/MyDataFixtures');
 
+Or you can load a set of fixtures from a file:
+
+    $loader->loadFromFile('/path/to/MyDataFixtures/MyFixture1.php');
+
 You can get the added fixtures using the getFixtures() method:
 
     $fixtures = $loader->getFixtures();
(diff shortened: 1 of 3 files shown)


lavoiesl:
Can we please get this merged? This is really annoying.

Even more infuriating, the documentation of the command wrongly says it supports specifying files: `The directory or file to load data fixtures from.`

lavoiesl:
testReferenceReconstruction error will be fixed by #177

PR #90, opened 2013-02-12
On doctrine:fixtures:load if data-fixtures cannot get a reference will f...

entering:
On doctrine:fixtures:load if data-fixtures cannot get a reference will fail with a undefined index.

I think this can be improve, to really pass the message undefined reference.

So I improved from this:
[ErrorException]
Notice: Undefined index: database_1_1 in /home/www/eb/platform/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php line 145

To:
[Exception]
Undefined reference database_1_1


lavoiesl:
Fixed by #179

PR #97, opened 2013-03-20
fix #96 refactor Loader to override hash method

### LINKED ISSUE #96: refactor Load to easier override get_class-hash method
hi there,

before i bootstrap everyhing an PR it:
for me the indexing with get_class in the loader does not work very well. I think you have a good reason to hash wich class-FQN, but i would love overload the loader, which is a little bit complicated

can I refactor the loader to use a new method `hashFixture()` in `hasFixture` and  `addFixture` to make the loader easier extendable?

best regards

pscheit:
okay it was that easy..

just a little bit refactoring to override the hash method (currently get_class) easier

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -116,7 +116,7 @@ class Loader
      */
     public function hasFixture($fixture)
     {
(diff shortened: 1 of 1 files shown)


lavoiesl:
Closing this PR since the milestone would be 2.0, please continue discussion in #96 and reopen if needed.

PR #106, opened 2013-07-24
Add a method to load fixtures from a single file

bobey:
I can't find a way to load fixtures from a given file through DoctrineFixturesBundle.
This is the first step to allow it with LoadDataFixturesDoctrineCommand.


lavoiesl:
Fixed by #79

PR #110, opened 2013-08-19
Allowing getReference to detect when the reference does not exist.

westinpigott:
If the reference requested does not exist, the $reference variable is NULL which cause the "$this->manager->getClassMetadata(get_class($reference));" line to attempt to get the metadata for Doctrine\Common\DataFixtures\ReferenceRepository (default because $referenceis NULL) rather than the class of the reference object.


lavoiesl:
Fixed by #179

PR #112, opened 2013-09-12
Avoid adding duplicate tables in the purge-list

Seldaek:
This can happen with table inheritance or something, I had many many duplicates in there.


lavoiesl:
Actually, it doesn’t to be needed anymore because of `getCommitOrder` calling `hasClass`.

PR #116, opened 2013-09-21
Fixing failing test cases and Loader issues.

tPl0ch:
Since I wanted to start developing on some feature, I was trying
to run the test suite which was impossible due to some autoload issues
as well as some programatic issues which were causing infinite
recursion in the DependentFixtureTest and Fatal Errors.

- [x] Adding 'Test.php' suffix to phpunit.xml.dist
- [x] Use require over require_once in tests/bootstrap.php
- [x] Fixing Loader to check for dependency class existance to not produce fatal error
- [x] Checking for parent and child class before adding dependency
- [x] Using namespaces starting from root in annotations for better IDE compatibility
- [ ] Fix ProxyReferenceRepositoryTest
- [x] Enable travis CI | Enable Scrutinizer CI

This eliminates the issues with testing described here:
https://github.com/doctrine/data-fixtures/pull/90

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: php
+
+php:
+  - 5.3
+  - 5.4
+  - 5.5
+
+before_script:
+  - composer selfupdate
+  - composer install --prefer-dist --dev
+
+script: phpunit --colors --coverage-text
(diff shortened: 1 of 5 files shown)


lavoiesl:
This PR is too big, it fixes too many things.

For the dependant fixtures, see my PR: #176, it does not duplicate the logic of checking for dependant classes.

lavoiesl:
@stof, it seems you are right, I went to the very first commit where that test was added (5bc3727) and the error is present.

lavoiesl:
Since this was covered in other PRs, I am closing this one.

PR #118, opened 2013-10-24
Changed the MongoDBPurger class to only clear collections

gerryvdm:
Changed the MongoDBPurger class to only clear collections, not remove them.

Instead of dropping all collections, we should remove all documents in a
collection instead. This avoids problems where indexes are lost while dropping
the collection, and collections not being recreated (with a schema create tool)
while loading fixtures.

Fixes #150

PR #123, opened 2013-11-27
Use purgeCollection method for drop separate collection

spolischook:


lavoiesl:
Why is that? Is it because you have problems where it drop your indices? If yes, #118 would fix it.

spolischook:
It's so old, so I don't remember why. Maybe for better testing. It can be closed anyway.

lavoiesl:
Just making sure.

Issue #17, opened 2011-06-12
Truncating table with foreign keys fails


ghost:
When I am calling symfony CLI command:

```
php app/console doctrine:fixtures:load
```

I get 

```
[PDOException] 

SQLSTATE[42000]: Syntax error or access violation: 1701 Cannot truncate a table referenced in a foreign key constraint (symfony.param_product, CONSTRAINT param_product_ibfk_1 FOREIGN KEY (param_val_id) REFERENCES symfony.param_value (id))
```

Table symfony.param_product is created by @ORM\JoinTable annotation (with @ORM\ManyToMany). I am using MySQL

This worked fine before this change: 91ff6ebbb781441c60782d90a4dd95482eeedf35

I thought it might be bug on my side (e.g. no onDelete options) but everything I tried failed.

As far as I learned e.g. from http://forums.asp.net/t/1283840.aspx/1?How+can+I+truncate+the+table+which+have+foreign+key+

```
 ON DELETE CASCADE is true only for deleting records and not for truncating tables.

 You have to drop the foreign key constraint from Child Table that references the Master Table to be truncated, then after only you are able to truncate the Master Table. 
```

So it seems to me the only solution is to use DELETE or drop the foregin keys first.

Can we revert the 91ff6ebbb781441c60782d90a4dd95482eeedf35 change to fix this problem? @beberlei?

sergiu-popa:
Do yourself a favor and create a script inside Symfony folder, like `load_fixtures`:
`
bin/console doc:sc:drop --force
bin/console doc:sc:cr
bin/console doc:fix:lo --no-interaction
`
Run `chmod 755 load_fixtures` and then `./load_fixtures.`

PR #127, opened 2014-01-15
Fix truncate on MySQL >= 5.5

### LINKED ISSUE #113: Support truncating tables with foreign keys
I had to override the loadFixtures method of the LiipFunctionalTestBundle with this stuff to avoid blowing up mysql:

```
    protected function loadFixtures(array $classNames, $omName = null, $registryName = 'doctrine', $purgeMode = null)
    {
        $conn = $this->getContainer()->get($registryName)->getManager()->getConnection();
        $db = $conn->getDriver()->getDatabasePlatform()->getName();
        if ($db === 'mysql') {
            $conn->exec('SET FOREIGN_KEY_CHECKS=0');
        }
        $res = parent::loadFixtures($classNames, $omName, $registryName, $purgeMode);
        if ($db === 'mysql') {
            $conn->exec('SET FOREIGN_KEY_CHECKS=1');
        }

        return $res;
    }
```

Now obviously it could be integrated in the bundle (cc @lsmith77) but ideally it would be part of this library itself so everyone benefits. Not sure how/where to best do this though, so I'm dropping this here hoping someone will pick it up :)

ddeboer:
When loading fixtures with purge mode truncate (`--purge-with-truncate` in the bundle's command) on MySQL >= 5.5, an error is thrown:

``` mysql
SQLSTATE[42000]: Syntax error or access violation: 
1701 Cannot truncate a table referenced in a foreign key constraint ...
```

It turns out that with MySQL 5.5, the `TRUNCATE` behaviour has changed. From the [MySQL docs](http://dev.mysql.com/doc/refman/5.5/en/truncate-table.html):

> TRUNCATE TABLE fails for an InnoDB table if there are any FOREIGN KEY constraints from other tables that reference the table. Foreign key constraints between columns of the same table are permitted.

With this PR foreign key checks are disabled before starting the purge, and re-enabled after, so purge made truncate works again. Fixes #113, #17.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/AbstractFixture.php
+++ b/lib/Doctrine/Common/DataFixtures/AbstractFixture.php
@@ -63,11 +63,14 @@ abstract class AbstractFixture implements SharedFixtureInterface
     /**
      * Set the reference entry identified by $name
      * and referenced to managed $object. If $name
-     * already is set, it overrides it
+     * already is set, it throws a 
+     * BadMethodCallException exception
      * 
      * @param string $name
      * @param object $object - managed object
      * @see Doctrine\Common\DataFixtures\ReferenceRepository::addReference
+     * @throws BadMethodCallException - if repository already has
+     *      a reference by $name
      * @return void
      */
     public function addReference($name, $object)
--- a/lib/Doctrine/Common/DataFixtures/FixtureInterface.php
+++ b/lib/Doctrine/Common/DataFixtures/FixtureInterface.php
@@ -33,5 +33,5 @@ interface FixtureInterface
      *
      * @param ObjectManager $manager
      */
-    function load(ObjectManager $manager);
+    public function load(ObjectManager $manager);
 }
--- a/lib/Doctrine/Common/DataFixtures/FixtureInterface.php
+++ b/lib/Doctrine/Common/DataFixtures/FixtureInterface.php
@@ -31,7 +31,7 @@ interface FixtureInterface
     /**
      * Load data fixtures with the passed EntityManager
      *
-     * @param Doctrine\Common\Persistence\ObjectManager $manager
+     * @param ObjectManager $manager
      */
     function load(ObjectManager $manager);
 }
--- a/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php
@@ -28,6 +28,7 @@ use Doctrine\Common\Util\ClassUtils;
  * Allow data fixture references and identities to be persisted when cached data fixtures
  * are pre-loaded, for example, by LiipFunctionalTestBundle\Test\WebTestCase loadFixtures().
  *
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
  * @author Anthon Pang <anthonp@nationalfibre.net>
  */
 class ProxyReferenceRepository extends ReferenceRepository
@@ -59,12 +60,13 @@ class ProxyReferenceRepository extends ReferenceRepository
      */
     public function serialize()
     {
+        $unitOfWork       = $this->getManager()->getUnitOfWork();
         $simpleReferences = array();
 
         foreach ($this->getReferences() as $name => $reference) {
             $className = $this->getRealClass(get_class($reference));
 
-            $simpleReferences[$name] = array($className, $reference->getId());
+            $simpleReferences[$name] = array($className, $this->getIdentifier($reference, $unitOfWork));
         }
 
         $serializedData = json_encode(array(
@@ -90,7 +92,7 @@ class ProxyReferenceRepository extends ReferenceRepository
                 $name,
                 $this->getManager()->getReference(
                     $proxyReference[0], // entity class name
-                    $proxyReference[1]  // id
+                    $proxyReference[1]  // identifiers
                 )
             );
         }
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\Common\DataFixtures\Purger;
 
+use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Internal\CommitOrderCalculator;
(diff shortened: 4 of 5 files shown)


iBasit:
+1

Soullivaneuh:
:+1: What is still needed to get it merged?

brunohanai:
I face this issue today and this workaround helped me: https://coderwall.com/p/staybw/workaround-for-1701-cannot-truncate-a-table-referenced-in-a-foreign-key-constraint-using-doctrine-fixtures-load-purge-with-truncate

jopais:
+1

famoser:
> I faced this issue today and this workaround helped me: https://coderwall.com/p/staybw/workaround-for-1701-cannot-truncate-a-table-referenced-in-a-foreign-key-constraint-using-doctrine-fixtures-load-purge-with-truncate

This did not work for me as I am using `DATABASE_URL` to configure the dbal, and as a consequence the suggested `driver_class` configuration is ignored. I hope it is not rude to post the full workaround I ended up using; I image most who have this problem end up here and would like to see one.

Caveat of this workaround: **The database platform has to be hard-coded**. This was acceptable for me in the dev environment.

## Step-by-Step
Find the platform you want to use as a parent class here: https://github.com/doctrine/dbal/tree/2.12.x/lib/Doctrine/DBAL/Platforms.
Then create you own platform, overriding the `getTruncateTableSQL` method with:

```php
# src/Doctrine/Platform.php
public function getTruncateTableSQL($tableName, $cascade = false)
{
    $truncateSql = parent::getTruncateTableSQL($tableName, $cascade);
    return 'SET foreign_key_checks = 0;'.$truncateSql.';SET foreign_key_checks = 1;';
}
```

Find the driver your want to use as a parent class here: https://github.com/doctrine/dbal/tree/2.12.x/lib/Doctrine/DBAL/Driver. 
Then create your own driver, overriding the `getDatabasePlatform` method with:

```php
# src/Doctrine/Driver.php
public function getDatabasePlatform()
{
    return new Platform();
}
```

Configure your new driver with the `driver_class` option. You additionally need to set `server_version` to some "old" version, else [`getDatabasePlatform()` is never called](https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php#L136). If you use a connection string, you need to remove the `schema` (like `mysql://`), else [the `driver_class` url is ignored](https://github.com/doctrine/dbal/blob/2.12.x/lib/Doctrine/DBAL/DriverManager.php#L473). 

```yaml
# config/packages/dev/doctrine.yaml
doctrine:
    dbal:
        driver_class: App\Doctrine\Driver
        server_version: 'mariadb-0.0.0' # nonexistant mariadb version so the platform provided by our driver is used
        url: '%env(trimschema:DATABASE_URL)%' # stripschema removes the schema from the DATABASE_URL
```

You can create `stripschema` [EnvVarProcessor](https://symfony.com/doc/current/configuration/env_var_processors.html) with:

```php
# src/Doctrine/TrimSchemaEnvVarProcessor.php
namespace App\Doctrine;

use Symfony\Component\DependencyInjection\EnvVarProcessorInterface;

class TrimSchemaEnvVarProcessor implements EnvVarProcessorInterface
{
    public function getEnv(string $prefix, string $name, \Closure $getEnv)
    {
        $env = $getEnv($name);

        return substr($env, strpos($env, '://') + 3);
    }

    public static function getProvidedTypes()
    {
        return [
            'trimschema' => 'string',
        ];
    }
}
```

## Full Driver & Platform example for MariaDb

```php
# src/Doctrine/Platform.php
namespace App\Doctrine;

use Doctrine\DBAL\Platforms\MySqlPlatform;

// cannot use MariaDb1027Platform as parent (as marked final), hence using its parent and then copying its methods
class Platform extends MySqlPlatform
{
    public function getTruncateTableSQL($tableName, $cascade = false)
    {
        $truncateSql = parent::getTruncateTableSQL($tableName, $cascade);

        return 'SET foreign_key_checks = 0;'.$truncateSql.';SET foreign_key_checks = 1;';
    }

    public function getJsonTypeDeclarationSQL(array $column): string
    {
        return 'LONGTEXT';
    }

    protected function getReservedKeywordsClass(): string
    {
        return MariaDb102Keywords::class;
    }

    protected function initializeDoctrineTypeMappings(): void
    {
        parent::initializeDoctrineTypeMappings();

        $this->doctrineTypeMapping['json'] = Types::JSON;
    }
}
```

```php
# src/Doctrine/Driver.php
namespace App\Doctrine;

use Doctrine\DBAL\Driver\PDOMySql;

class Driver extends PDOMySql\Driver
{
    public function getDatabasePlatform()
    {
        return new Platform();
    }
}
```
Commit implementing this workaround in a symfony project: https://github.com/mangelio/web/commit/64a171e97ec00a13e66f2cf78d90117319ff99ef

mannion007:
I wrote a small decorator to wrap the existing ORM purger with the additional behaviour required. 

Perhaps it could also verify that the platform is MySQL and give a useful error if it isn't.

```php
/**.
 * @see https://github.com/doctrine/DoctrineFixturesBundle/issues/50
 */
final class ForeignKeyDisablingOrmPurgerDecorator implements ORMPurgerInterface
{
    public function __construct(
        private readonly ORMPurgerInterface $delegate,
        private readonly Connection $connection,
    ) {}

    public function purge()
    {
        $this->connection->executeStatement('SET FOREIGN_KEY_CHECKS=0');
        $this->delegate->purge();
        $this->connection->executeStatement('SET FOREIGN_KEY_CHECKS=1');
    }

    public function setEntityManager(EntityManagerInterface $em)
    {
        $this->delegate->setEntityManager($em);
    }
}
```

Following the docs, add a factory for that decorator

```php
final class ForeignKeyDisablingOrmPurgerDecoratorFactory implements PurgerFactory
{
    public function createForEntityManager(
        ?string $emName,
        EntityManagerInterface $em,
        array $excluded = [],
        bool $purgeWithTruncate = false,
    ): PurgerInterface {
        $ormPurger = new ORMPurger($em, $excluded);
        $ormPurger->setPurgeMode($purgeWithTruncate ? ORMPurger::PURGE_MODE_TRUNCATE : ORMPurger::PURGE_MODE_DELETE);

        return new ForeignKeyDisablingOrmPurgerDecorator($ormPurger, $em->getConnection());
    }
}
```

 [Register it](https://symfony.com/bundles/DoctrineFixturesBundle/current/index.html#specifying-purging-behavior) and then specify it by alias when loading fixtures:

```
bin/console doctrine:fixtures:load --purger=foreign_key_disabling
```

PR #131, opened 2014-01-24
Allow dependent fixtures to reference other types of fixture without triggering an error.

r00ps:
When mixing ordered and dependent fixtures, if a dependent fixture references an ordered fixture in getDependencies it triggers an error during loading. This is because the ordered fixtures are not included in the list of classes that need to be sequenced, triggering a Notice error when using the class name as an array index.

This  fix checks to see if the class name exists in the array before checking if it has been sequenced already.


lavoiesl:
Can you rebase and provide a test for this? I am not sure it is still needed.

PR #132, opened 2014-01-29
allow the use of multiple transactions - one per file

### LINKED ISSUE #126: ORMExecutor wraps all fixtures in a single transaction
Hello,

  So I have a situation where the current executor can't load my fixtures. I have three fixture files. The third one loaded has an entity that has a custom id generator. The id generator is based of an incrementing value in objects in the first fixture.

The current executor is

```
{
        $executor = $this;
        $this->em->transactional(function(EntityManager $em) use ($executor, $fixtures, $append) {
            if ($append === false) {
                $executor->purge();
            }
            }
        });
}
```

I've used a couple of different machines with this, once got a segfault other times got other php fatal errors - don't have all of them at the moment but can dig up specific issues if necessary. In any case I'm wondering why the entire fixture loading is in one transaction. When I change that execute to

```
{
    if ($append === false) {
        $this->purge();
    }
        $this->em->beginTransaction();
        $this->load($this->em, $fixture);
        $this->em->commit();
    }
}
```

My fixtures load. 

So I'm not sure why the entire process needs to be one transaction as opposed to a transaction per fixture file. Thoughts?

gnat42:
This allows the user to select whether they want a single transaction for all fixture files or one transaction per fixture file. This fixes a bug where I had a custom id generator that queries another table and increments an id there within its own transaction. It causes a segfault.

Fixes #126, #42

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -132,11 +132,13 @@ class ORMPurger implements PurgerInterface
             $orderedTables[] = $this->getTableName($class, $platform);
         }
 
+        $connection = $this->em->getConnection();
         foreach($orderedTables as $tbl) {
             if ($this->purgeMode === self::PURGE_MODE_DELETE) {
-                $this->em->getConnection()->executeUpdate("DELETE FROM " . $tbl);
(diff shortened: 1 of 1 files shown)


lavoiesl:
I would tend to agree with this, could you rebase your commit?

gnat42:
@lavoiesl So it took me long enough but I've provided a rebased version in #209

PR #139, opened 2014-03-30
Fix Quoting of "joinTable" table names in ORMPurger

elkangaroo:
Using database or table names that need quotation in @JoinTable annotations lead to errors in ORMPurger when using data-fixtures. @Table annotions are correctly quoted.

This commit fixes the issue.

Example:
```php
/**
 * @JoinTable(name="`my-database.table")
 */
private $relation;
```

... would create the following query, which fails (MySQL):
```sql
DELETE FROM my-database.table;
```

... after the fix, creates the following query, which succeeds (MySQL):
```sql
DELETE FROM `my-database`.table;
```


lavoiesl:
Closing as superseeded

PR #141, opened 2014-04-22
Add support for array of objects as reference

flip111:
Could not start test from working test code, see https://github.com/doctrine/data-fixtures/issues/142

Why is this useful? Consider model relations:
`A OneToMany B OneToMany C` (a tree)
When creating the entities for `B`, it's easy to grab that one `A`. However when creating fixtures for `C` there are now multiple `B`'s, they clearly belong together (are part of the same A), but they can't be grouped. The only alternative is to do `addReference('B1', $Bs[0]); addReference('B2', $Bs[1]);` etc

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
@@ -70,14 +70,23 @@ class ReferenceRepository
      * @param object $reference Reference object
      * @param object $uow       Unit of work
      *
-     * @return mixed
+     * @return array
      */
     protected function getIdentifier($reference, $uow)
     {
+        // In case Reference is not yet managed in UnitOfWork
+        if ( ! $uow->isInIdentityMap($reference)) {
+            $class = $this->manager->getClassMetadata(get_class($reference));
+
+            return $class->getIdentifierValues($reference);
+        }
+
+        // Dealing with ORM UnitOfWork
         if (method_exists($uow, 'getEntityIdentifier')) {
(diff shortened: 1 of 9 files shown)


lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -98,7 +98,7 @@ protected function getIdentifier($reference, $uow)
      * @param string $name
      * @param object $reference
      */
-    public function setReference($name, $reference)
+    public function setReference($name, object $reference)
```
stof: `object` is not a valid typehint in PHP

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -109,6 +109,22 @@ public function setReference($name, $reference)
     }
 
     /**
+     * Convenience function to set multiple references at once.
```
stof, line 112: setting multiple references for a single name does not make sense

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -129,17 +145,53 @@ public function setReferenceIdentity($name, $identity)
...
+     */
+    public function getReference($name)
+    {
+        $references = $this->references[$name];
+
+        if (! is_array($references)) {
+            return $this->_loadReference($name, $references);
+        }
+
+        $return = [];
```
stof: invalid PHP 5.3 code

PR #143, opened 2014-04-22
Update bootstrap.php

flip111:
Composer does not give back `$loader` object when using `require_once` (returns `true` instead). Use require to get object.

The testing bootstrap does not support being started from within the vendor folder. Only run from the `src` folder is supported. The bootstrap could be more adaptive like this https://github.com/doctrine/doctrine2/blob/master/tests/Doctrine/Tests/TestInit.php

(overall i noticed the testing bootstrap is inconsistent throughout doctrine libraries)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ matrix:
   allow_failures:
     - php: hhvm
 
-before_script: composer --dev --prefer-source install
+before_script:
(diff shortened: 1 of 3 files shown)


lavoiesl:
Removed the Travis self-update since the version is long updated.

PR #144, opened 2014-04-24
allow access to the referenceRepository

flip111:
Unless you want to put a decorator for every method, allowing direct access to the ReferenceRepository is a good idea. For example right now you can not call https://github.com/doctrine/data-fixtures/blob/master/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php#L216


lavoiesl:
It seems to me that it should either be all decorated or not at all. `getReferences` is understandable, what other method would be needed?

flip111:
I'm totally lost on this PR. Github shows a lot of commits by other people i don't know why. And also i totally forgot why i needed this in the first place.

lavoiesl:
Rebase your branch on top of master

flip111:
@lavoiesl as per your previous comment it seems that the PR is still open for debate. Actually i don't know if it would be worth to "prepare it for merging" just yet .. also because i really added one very simple method https://github.com/flip111/data-fixtures/commit/6385ae03ab034525053cc5c59c67992b85de9e0e

lavoiesl:
I would not merge it without more discussion, but it still needs to be rebased if we want to look at it.

flip111:
I will have to drop my support on this PR. Sorry i am not in the same position as i was on 24 apr 2014.

lavoiesl:
No worries.

PR #145, opened 2014-04-24
Fill simpleReferences with IDs using identities

webcome:
The $simpleReferences don't containt the array with the IDs in some cases but they are in the identities. Then we can use the identities to fill the $simpleReferences.


lavoiesl:
Can you provide a test case for this?

PR #149, opened 2014-05-17
escape table name on purge

abimaelmartell:
Fix quoting of table name on ORMPurger delete query.


lavoiesl:
Is it me or this is useless since the table name is already quoted in https://github.com/abimaelmartell/data-fixtures/blob/fix/escape-table-name-on-purge/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php#L130 ?

lavoiesl:
I just tested and the quoting is already done.

Issue #135, opened 2014-03-10
Compound ids trip up the ProxyReferenceRepository


Seldaek:
I have a class that has two identifier fields (a numeric id and a user ManyToOne) so that every user has their own sequence of numeric ids. This causes failures together with the liip functional test bundle when reloading a sqlite DB from cache I believe.

Here is a stack trace:

```
Array to string conversion

vendor\doctrine\orm\lib\Doctrine\ORM\UnitOfWork.php:2899
vendor\doctrine\orm\lib\Doctrine\ORM\EntityManager.php:512
app\cache\test\jms_diextra\doctrine\EntityManager_531d8d621ddd2.php:240
vendor\doctrine\data-fixtures\lib\Doctrine\Common\DataFixtures\ProxyReferenceRepository.php:97
vendor\doctrine\data-fixtures\lib\Doctrine\Common\DataFixtures\ProxyReferenceRepository.php:123 vendor\liip\functional-test-bundle\Liip\FunctionalTestBundle\Test\WebTestCase.php:274
```

The ProxyReferenceRepository's serialize and unserialize methods don't deal with non-numeric ids it seems.

Relates to https://github.com/doctrine/data-fixtures/commit/8ffac1c63f34124f693b93889fa32f4036eb241b kinda.

If I dump the $proxyReference at [line 91](https://github.com/doctrine/data-fixtures/blob/master/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php#L91) it is like this:

```
array(2) {
  [0] =>
  string(32) "Entity\Object"
  [1] =>
  array(2) {
    'id' =>
    int(1)
    'parent' =>
    array(3) {
      '__initializer__' =>
      array(0) {
        ...
      }
      '__cloner__' =>
      array(0) {
        ...
      }
      '__isInitialized__' =>
      bool(false)
    }
  }
}
```

/cc @guilhermeblanco

JustBlackBird:
Is there any progress on the issue?

vviippola:
+1

paulinevos:
+1, March 2018 and still having this problem. We can't use the ORM as intended now because tests fail upon loading fixtures.

Ocramius:
@JustBlackBird @vviippola @paulinevos please help out on https://github.com/doctrine/data-fixtures/pull/167

PR #167, opened 2015-01-12
Fix composite key serialization

ddeboer:
Currently, calling `unserialize()` with some data previously returned from `serialize()` produces:

``` php
Array to string conversion

/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:2913
```

This is caused by `serialize()` incorrectly serializing entities with primary keys (on doctrine/orm 2.4.):

``` php
"ref1":["Entity\\MyEntity",{"user":{"__initializer__":{},"__cloner__":{},"__isInitialized__":false},"somethingElse":{"__initializer__":null,"__cloner__":null,"__isInitialized__":true}}]
```

This seems to be caused by the identifier lookup for `$simpleReferences` not working properly. For each object constituting the composite key, it returns the object itself instead of its id. E.g.:

``` php
array(2) {
  [0] =>
  string(38) "Entity\MyEntity"
  [1] => 
    'user' =>
    class Proxies\__CG__\Entity\User#4711 (25) {
        // ...
    },
    'somethingElse' =>
    class Proxies\__CG__\Entity\SomethingElse#1234 (25) {
        // ...
    }
```

By adding a `getReference()` the identifier lookup works correctly. It now looks at `$this->identities` first, which holds the correct composite identities. Of course, the extra call does have a slight performance impact:

``` php
array(2) {
  [0] =>
  string(38) "Entity\MyEntity"
  [1] =>
  array(3) {
    'user' =>
    int(4)
    'somethingElse' =>
    int(1)
  }
}
```


ddeboer:
This may be a (partial) fix for #135.

lavoiesl:
Can you provide tests for that?

lavoiesl:
@ddeboer, also, please rebase your commits.

lavoiesl:
@ddeboer, I would like to be able to release this PR in 1.1 by tomorrow, do you think you would be able to care of it?

ddeboer:
Rebased. Now working out how I can test this.

tests/Doctrine/Tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
```diff
@@ -22,7 +22,6 @@
 use Doctrine\Common\DataFixtures\ProxyReferenceRepository;
 use Doctrine\Common\DataFixtures\Event\Listener\ORMReferenceListener;
 use Doctrine\ORM\Tools\SchemaTool;
-use Doctrine\ORM\Proxy\Proxy;
```
ddeboer: Was unused.

ddeboer:
Done.

lavoiesl:
Awesome! Thank you.

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -77,8 +77,24 @@ protected function getIdentifier($reference, $uow)
...
+            // https://github.com/doctrine/data-fixtures/issues/167
+            foreach ($values as $key => $value) {
+                if (!is_scalar($value)) {
+                    // To prevent "Array to string conversion" in
+                    // UnitOfWork::tryGetById, we can only return a single
+                    // identifier. This does mean that related entities that
+                    // themselves have a composite primary key are unsupported.
+                    $proxyId = $this->getIdentifier($value, $uow);
+                    $keys = array_keys($proxyId);
+                    $values[$key] = $proxyId[$keys[0]];
```
lavoiesl: Sorry if I’m misunderstanding something, but wouldn’t this create ID collisions? Why not use `implode('|', $keys)`?

ddeboer: In most cases, where the related entity has a single (not composite) primary key, `count($proxyId) === 1`.

Imploding the keys is an option, but unfortunately won’t help much. When the references are unserialized, references are [retrieved from the entity manager](https://github.com/doctrine/data-fixtures/blob/master/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php#L93). However, Doctrine does not support something like:

```php
$roleRef = $em->getReference(self::TEST_ENTITY_USER_ROLE, [
    'role' => 1,
    'user' => [
        'id' => 1,
        'code' => '007'
    ]
]);

```

lavoiesl:
After having carefully considered this PR and asked other guys on the Doctrine team, we won’t be accepting it as is. Before releasing support for composite keys, it will need to support all cases and/or have tests for all those.

I will reschedule this PR for 1.2 and will come back when I have a better understanding.

Thanks again.

ddeboer:
I too prefer full support for nested composite keys, but as this requires changes in Doctrine itself I think starting simple, with support for unnested primary keys, is the way to go. Why does the solution have to tackle all cases at once? Why not work on a solution iteratively?

PR #170, opened 2015-02-13
Verify if reference is null (and skip all NULLs) while serializing references

kniziol:
Why?
1. It's required to avoid error `spl_object_hash() expects parameter 1 to be object, null given` while running functional tests
2. This problem occurs when some of loaded entities are deleted. After that there are references that points to `null`.

Example of stack trace:

```
/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:1540
/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php:78
/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php:79
/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php:144
/vendor/liip/functional-test-bundle/Liip/FunctionalTestBundle/Test/WebTestCase.php:337
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: php
+
+php:
+  - 5.3
+  - 5.4
+  - 5.5
+  - 5.6
+  - hhvm
+
+matrix:
+  allow_failures:
+    - php: hhvm
+
+before_script: composer --dev --prefer-source install
+
+script: ./vendor/bin/phpunit -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ matrix:
 
 before_script: composer --dev --prefer-source install
 
-script: ./vendor/bin/phpunit -v
+script: phpunit -v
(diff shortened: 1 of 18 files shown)


jwage:
@kniziol Can you demonstrate with a test?

kniziol:
@jwage This problem occurs when you're using ```liip/functional-test-bundle``` repo and:
1. You are loading DataFixtures using ```Liip\FunctionalTestBundle\Test\WebTestCase::loadFixtures()``` method
2. Some of the references to the fixtures are ```null```, because they were deleted
3. DataFixtures are serialized using ```$executor->getReferenceRepository()->save($backup)``` method
4. And here is the essence of problem: ```Doctrine\Common\DataFixtures\ProxyReferenceRepository::serialize()``` method doesn't verify provided ```$reference```

jwage:
@kniziol Thanks for the explanation. Can you rebase the PR and add a test case for the fix?

PR #175, opened 2015-03-19
Load fixture from file

fnash:


jwage:
This should be fixed with #79

PR #177, opened 2015-03-22
Fixed test for instanceof Doctrine\ORM\Proxy\Proxy

lavoiesl:
As @stof said:
Getting the reference before the EntityManager is cleared will not
return a proxy instance as the manager already contains a managed
instance.

This test always failed, even where it was first introduced: 5bc3727

Fixes a task in #116

### DIFF
--- a/tests/Doctrine/Tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
@@ -113,7 +113,9 @@ class ProxyReferenceRepositoryTest extends BaseTest
 
         $ref = $proxyReferenceRepository->getReference('admin-role');
 
-        $this->assertInstanceOf('Doctrine\ORM\Proxy\Proxy', $ref);
+        // before clearing, the reference is not yet a proxy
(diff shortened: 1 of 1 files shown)


jwage:
:+1:

guilhermeblanco:
@lavoiesl @jwage I am very concerned about the potential impact over these new changes into the original purpose of the ProxyReferenceRepository.
When it was originally created, we (InstaClick team) was trying to optimize DB fetching by storing references in a serialized file and reusing on consecutive imports.
Let me explain the scenario:

I have a bunch of DB tables which are imported normally through the schema and also import initial base data required for the app to work. Under SQLite, we cache the generated schema in a separate file and reuse for every functional test. Now, we need to access somehow these initial base data through their references, but on second functional test, it would just copy over the .db file and nothing would be fetched. That's where ProxyReferenceRepository kicks in.
In this circumstance, we copy the data over and you're still allowed to use the originally set base data through the references, but you'd always retrieve proxies (not fully loaded entities).

Does that make sense now? This behavior needs to be preserved, otherwise many people that use either LiipFunctionalTestBundle or ICBaseTestBundle would suddenly have all their functional tests to not work as soon as data-fixtures library gets updated.

jwage:
@guilhermeblanco which changes are you referring to? As far as I know the changes in this pull request are to fix a test that has always been failing.

PR #180, opened 2015-03-22
Fix Quoting of "joinTable" table names in ORMPurger

lavoiesl:
Fixes #139


Ocramius:
@lavoiesl this needs an additional test case as well

lavoiesl:
True that.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -185,17 +185,22 @@ private function getCommitOrder(EntityManagerInterface $em, array $classes)
         return $calc->getCommitOrder();
     }
 
-    private function getAssociationTables(array $classes)
+    /**
+     * @param array $classes
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     * @return array
+     */
+    private function getAssociationTables(array $classes, $platform)
```
stof: the argument should be typehinted

lavoiesl: fixed

Issue #140, opened 2014-04-07
Idea: grouping fixtures


rodrigorigotti:
I use DataFixtures for several reasons, but they usually group into one of the following cases: a) creating data that is required for my application, when a fresh new installation is made, and b) creating dummy data to simulate my application running with a lot of data.
I think we should, somehow, group the fixtures classes. An example could be something like:

```
namespace MyDataFixtures;

use Doctrine\Common\DataFixtures\AbstractFixture;
use Doctrine\Common\DataFixtures\GroupedFixtureInterface;
use Doctrine\Common\Persistence\ObjectManager;

class LoadUserData extends AbstractFixture implements GroupedFixtureInterface
{
    public function load(ObjectManager $manager)
    {
    }

    public function getGroup()
    {
        return "basic_data";
    }
}

class LoadCheckinData extends AbstractFixture implements GroupedFixtureInterface
{
    public function load(ObjectManager $manager)
    {
    }

    public function getGroup()
    {
        return "extended_data";
    }
}
```

... then Doctrine would just load the fixtures from the group the user has requested (when a group is not provided, Doctrine loads all the fixtures).

mssimi:
Hey, what does  "pretty close" means? Since it has been 3 years.

Ocramius:
data-fixtures 2.x never happened, as @guilhermeblanco got side-tracked. Won't happen for a while either, sorry.

mssimi:
Ok, thx for info.

PR #184, opened 2015-03-27
Don't purge readonly Entities

elkangaroo:
The ORMPurger currently tries to truncate database views (which obviously fails), even though the entity is marked as "readonly" using following ORM annotation:
```php
/**
 * @ORM\Entity(readOnly=true)
 * @ORM\Table(name="foo")
 */
```

This commit skips readonly entities in ORMPurger.


stof:
readonly in Doctrine ORM is not meant to cover views, but to cover immutable entities (the name is weird, I know). Doctrine ORM still allows to remove such an entity. What it does not allow is updating them (it does not track them for changes)

stof:
and btw, it also allows to insert such entities

lavoiesl:
I don’t think it is a good idea to generalize that readOnly entities should not be purge. Selective purging seems a better idea. See #44 and #140.

lavoiesl:
P.s.: This would need tests.

stof:
I'm closing this PR as merging it is impossible. It relies on a wrong understanding of the Doctrine feature.

PR #185, opened 2015-03-27
Revert "Use EntityManagerInterface instead of EntityManager"

### LINKED ISSUE #183: Compatibility break in 1.0.1
Hello,

I noticed a huge BCBreak in the patch version 1.0.1 here: https://github.com/doctrine/data-fixtures/commit/48ec2812ab8d3c044f975b69a0e834c59b0a4a57#diff-bcb90b22d8706b63fa51389ffc49ddfbR22

This part is not compatible with doctrine 2.3 (that is the default version with [the good old 2.3 standard symfony version](https://github.com/symfony/symfony-standard/blob/v2.3.11/composer.json)).

You can maybe add a warning somewhere... ?

lavoiesl:
Fixes #183

Reverts #166, it is causing the BC break, it will be added to `v1.1` instead.

Revert "Use EntityManagerInterface instead of EntityManager (to support decorators, etc)"
Revert "Upgrade doctrine/orm to ~2.4"
This reverts commit 0c96db584a4ccb7cfa4f68d8ee435a3a6f387e81.
This reverts commit 48ec2812ab8d3c044f975b69a0e834c59b0a4a57.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "doctrine/common": "~2.2"
     },
     "require-dev": {
-        "doctrine/orm": "~2.4"
+        "doctrine/orm": "~2.2"
     },
     "suggest": {
         "doctrine/orm": "For loading ORM fixtures",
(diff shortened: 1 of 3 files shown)


lavoiesl:
Wrong branch

PR #187, opened 2015-03-27
Added a conflict on doctrim/orm < 2.4

lavoiesl:
`EntityManagerInterface` is only available in 2.4. Since we do not hard depend on `doctrine/orm`, add it as conflict instead.

Related to #166 #183 #186


lavoiesl:
Merged in 1.1 branch

PR #165, opened 2014-12-12
Hotfix for #164: PHPCRLoader should be transactional

Ocramius:
Ping @electricmaxxx

See #164 
see doctrine/phpcr-odm#582

PR #190, opened 2015-03-30
handle older versions of phpcr-odm

dbu:
phpcr-odm only got transactional added in a recent version. we get  https://github.com/symfony-cmf/symfony-cmf/issues/217

the alternative would be to declare a `conflicts` in composer.json to force using a version of phpcr-odm that has transactional.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Executor/PHPCRExecutor.php
+++ b/lib/Doctrine/Common/DataFixtures/Executor/PHPCRExecutor.php
@@ -60,7 +60,7 @@ class PHPCRExecutor extends AbstractExecutor
     {
         $that = $this;
 
-        $this->dm->transactional(function ($dm) use ($append, $that, $fixtures) {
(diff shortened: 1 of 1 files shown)


lavoiesl:
ping @Ocramius

lavoiesl:
Related to #165

Ocramius:
:+1: 

Also not really testable :-(

dbu:
true that. we could mock the DocumentManager without transactional to test it...

lavoiesl:
@dbu I don’t really see the point, it’s fine like this. Also, the test failing would mean a fatal error.

dbu:
well rather a fatal error in test than for end users :-)

but i am fine if we can merge like this and tag a patch release for it soon. people following the cmf tutorials currently end up with fatal errors until 1.1.1 is released.

lavoiesl:
Released as 1.1.1, thanks.

dbu:
great, thanks you!

PR #197, opened 2015-06-26
fix DELETE sql query

DenisMedved:
Hello. 

Fix for sql error if table name is "order"

````sql                                                                                                                                                               
  An exception occurred while executing 'DELETE FROM order':                                                                                                                                                         
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'order' at line 1  
```

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -131,11 +131,13 @@ class ORMPurger implements PurgerInterface
             $orderedTables[] = $class->getQuotedTableName($platform);
         }
 
+        $connection = $this->em->getConnection();
         foreach($orderedTables as $tbl) {
             if ($this->purgeMode === self::PURGE_MODE_DELETE) {
-                $this->em->getConnection()->executeUpdate('DELETE FROM `' . $tbl . '`');
+                $tbl = $connection->quoteIdentifier($tbl);
+                $connection->executeUpdate('DELETE FROM ' . $tbl );
             } else {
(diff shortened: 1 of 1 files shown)


KingCrunch:
This only works for MySQL

BuCeFaL:
Ok, thanks you for quick response.

KingCrunch:
Hi @BuCeFaL 

Me again :smile: There should be something like `$connection->quoteIdentifier()`, which should (you may guess) quote the identifier according the current connection. I cannot proof the correct method names right now though.

BuCeFaL:
Fix it with quoteIdentifier

stof:
this change is weird. the table names are already retrieved using ``getQuotedTableName``.

I think the issue is that your ORM mapping was not telling the ORM to use a quoted identifier.

Forcing quoting here causes issues in 2 cases:
- if the mapping already configures quoting properly, you are not quoting it twice
- if the identifier is not configured to be quoted, you are now forcing to be quoted, and this changes the behavior of the query, potentially making it use a different table than the one used by the ORM (which would not quote the identifier in this case): http://www.alberton.info/dbms_identifiers_and_case_sensitivity.html#.Veco4bOaClN

DenisMedved:
I think, better solution is left all as was. Anyway it is a very bad idea use reserved keywords in table names. I revert my "force quote".

maryo:
It should also be reverted here since it is a bug. It causes double-quoting. If you want your table to be escaped, use backtics in table name definition like `@ORM\Table(name="`order`")`

PR #198, opened 2015-07-08
Demonstrate undefined index notice when using fixture dependent on ordered fixture.

brabeji:
This demonstrates that undefined index notice gets somehow suppressed during standalone tests. That notice stops execution when running doctrine:fixtures:load which is a problem. Sorry about the whitespace formatting, relevant change is at the end of Loader.php diff.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -48,7 +48,7 @@ class Loader
      * @var boolean
      */
     private $orderFixturesByNumber = false;
-    
+
     /**
      * Determines if we must order fixtures by its dependencies
      *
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/DataFixtures/Loader.php
```diff
@@ -322,6 +322,10 @@ private function getUnsequencedClasses($sequences, $classes = null)
         }
 
         foreach ($classes as $class) {
+            if(!isset($sequences[$class])) {
+                dump('UNDEFINED INDEX');
+                die;
```
stof: this should be removed, as it prevents merging the new test in the existing testsuite

brabeji: Unfortunately, without this check the notice is suppressed and the test passes. I am not sure why. This is rather demo than proper PR.

PR #201, opened 2015-08-04
Use UnitOfWork->contains() instead of isInIdentityMap when Phpcr document

aramalipoor:
Unfortunately `isInIdentityMap` does not exist in Phpcr's DocumentManager which is raising issues when for example we're trying to `setReference` a document.

I'm not an expert in Doctrine stuff so I'm not sure if this is the right way to do it.

Related issue https://github.com/doctrine/phpcr-odm/issues/492

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
@@ -20,6 +20,7 @@
 namespace Doctrine\Common\DataFixtures;
 
 use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\ODM\PHPCR\DocumentManager as PhpcrDocumentManager;
 
 /**
  * ReferenceRepository class manages references for
@@ -75,7 +76,7 @@ class ReferenceRepository
     protected function getIdentifier($reference, $uow)
     {
         // In case Reference is not yet managed in UnitOfWork
-        if ( ! $uow->isInIdentityMap($reference)) {
+        if ( ! $this->hasIdentifier($reference)) {
             $class = $this->manager->getClassMetadata(get_class($reference));
 
             return $class->getIdentifierValues($reference);
@@ -101,9 +102,10 @@ class ReferenceRepository
     public function setReference($name, $reference)
     {
         $this->references[$name] = $reference;
-        // in case if reference is set after flush, store its identity
-        $uow = $this->manager->getUnitOfWork();
-        if ($uow->isInIdentityMap($reference)) {
+
+        if ($this->hasIdentifier($reference)) {
+            // in case if reference is set after flush, store its identity
+            $uow = $this->manager->getUnitOfWork();
             $this->identities[$name] = $this->getIdentifier($reference, $uow);
         }
     }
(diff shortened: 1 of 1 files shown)


stof:
@lsmith77 @dbu can you check this ? Will it work properly when setting a reference before flushing ?

lsmith77:
maybe it makes more sense to use ``contains()`` on the ObjectManager:
https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Persistence/ObjectManager.php#L168

https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/EntityManager.php#L708

stof:
@lsmith77 won't work on the ORM, because the manager can contain the object without knowing its id if the object is scheduled for insertion and the id is an auto-increment in the DB, which will be known only after the flush

lsmith77:
so you need this logic?

so this is the logic you need?
https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/UnitOfWork.php#L1594

then probably this method is better:
https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/UnitOfWork.php#L3091

stof:
@lsmith77 this looks weird. the method you are suggesting expects an id as argument, while what we are trying to do is precisely to retrieve the id of the object (but only when it is available)

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -101,10 +102,18 @@ protected function getIdentifier($reference, $uow)
...
         $this->references[$name] = $reference;
+
         // in case if reference is set after flush, store its identity
         $uow = $this->manager->getUnitOfWork();
-        if ($uow->isInIdentityMap($reference)) {
-            $this->identities[$name] = $this->getIdentifier($reference, $uow);
+
+        if ($this->manager instanceof PhpcrDocumentManager) {
+            if ($uow->contains($reference)) {
+                $this->identities[$name] = $this->getIdentifier($reference, $uow);
```
stof: rather than duplicating all the logic getting the identity, you should change only the detection of the object, not the whole logic

stof:
There are a bunch of other usages of ``isInIdentityMap`` in this class, so the PR is incomplete

aramalipoor:
@stof I've created a helper method to use in every usage. What do you think?

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -158,8 +162,8 @@ public function getReference($name)
 
         $reference = $this->references[$name];
         $meta = $this->manager->getClassMetadata(get_class($reference));
-        $uow = $this->manager->getUnitOfWork();
-        if (!$uow->isInIdentityMap($reference) && isset($this->identities[$name])) {
+
+        if (!$this->hasIdentifier($reference) && isset($this->identities[$name])) {
```
stof: I think this case could actually use ``$this->manager->contains()`` as we want to determine whether the object is managed

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -101,13 +102,16 @@ protected function getIdentifier($reference, $uow)
...
+
+        if ($this->hasIdentifier($reference)) {
+            // in case if reference is set after flush, store its identity
+            $uow = $this->manager->getUnitOfWork();
             $this->identities[$name] = $this->getIdentifier($reference, $uow);
         }
     }
 
+
+
```
stof: should be reverted

aramalipoor:
@stof Done.

PR #192, opened 2015-04-08
Use QuoteStrategy from Configuration to get table names

plfort:
Use QuoteStrategy to get table names
See #191

Issue #191, opened 2015-04-08
Table schema is not taken in account when purging since doctrine 2.5


plfort:
Hello,
Since I upgraded doctrine orm to 2.5, doctrine fixtures can't purge databse with table with specific schema (defined with @Table(name="hello",schema="mySchema") )
I think we should use the DefaultQuoteStrategy instead of getQuotedTableName in Purger

getQuotedTableName in ClassMetadaInfo is deprecated since 2.3 and does not take in account the "schema" attribute in metadata.

anthonyherve:
Anther idea to resolve this problem : we can use method getSequencePrefix instead of getQuotedTableName

awildeep:
Confirmed, I applied a hack to load my fixtures, but it is not a portable one so I won't bother to post it here.  Perhaps there should be a getFullyQualifiedTableName() method?  Something that could be driven by the platform.

jfitzgeraldgroup:
We also have an issue with this as well. We are utilizing schemas in Postgresql and this is causing our Entities to break every time we run the Symfony console command to load fixtures. 

The documentation shows that we should use... 

    @Table(name="tblName",schema="schName") 

but this does not work. We either have to put our schema name in the 'name' field like so 

    @Table(name="schName.tblName") 

which causes other issues in Symfony 2.7, or hack ORMPurger.php every time we do a composer update. Very frustrating!

awildeep:
The correct format is:

    @Table(name="tblName",schema="schName") 

Even for Doctrine 2.6.  My guess is that the fixture loader was modified by a user that either does not use schemas, or doesn't have access to them to test properly.

BPScott:
This is fixed by #192

PR #202, opened 2015-08-07
Fix of table names in ORMPurger

fon60:
In new ORM schame names are not part of table name. Best solution is to use Quiting strategy which is platform dependent.


fon60:
#191 Resolves Table names. Doesn't touch joinTable names, because they have its own solution.

alexsegura:
This is a duplicate of https://github.com/doctrine/data-fixtures/pull/192

PR #204, opened 2015-08-20
Support schema attribute everywhere.

alexsegura:
Just to see if you think this is interesting. 
The same thing is done in the `getAssociationTables` method, and I think this doesn't break anything. 

Please don't merge yet, I still need to write a test case for this, but the `ORMPurgerTest` class is somewhat "hacky", I'm trying to rewrite it to avoid testing private methods.


alexsegura:
Closing it as this is a duplicate of https://github.com/doctrine/data-fixtures/pull/192 and https://github.com/doctrine/data-fixtures/pull/202

PR #209, opened 2015-10-24
allow the use of multiple transactions - one per file

### LINKED ISSUE #126: ORMExecutor wraps all fixtures in a single transaction

gnat42:
This allows the user to select whether they want a single transaction for all fixture files or one transaction per fixture file. This fixes a bug where I had a custom id generator that queries another table and increments an id there within its own transaction. It causes a segfault.

Fixes #126, #42 
This is the rebased version of PR #132 

That first PR was done prior to me really getting github PR workflow, and the attempted rebase failed. So here's the patch against latest master.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
+++ b/lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
@@ -141,6 +141,7 @@ abstract class AbstractExecutor
      *
      * @param array $fixtures Array of fixtures to execute.
      * @param boolean $append Whether to append the data fixtures or purge the database before loading.
+     * @param boolean $single_transaction Whether to use a single transaction when loading fixtures
      */
-    abstract public function execute(array $fixtures, $append = false);
+    abstract public function execute(array $fixtures, $append = false, $single_transaction = true);
 }
--- a/lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
+++ b/lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
@@ -77,16 +77,27 @@ class ORMExecutor extends AbstractExecutor
     }
 
     /** @inheritDoc */
-    public function execute(array $fixtures, $append = false)
+    public function execute(array $fixtures, $append = false, $single_transaction = true)
     {
-        $executor = $this;
-        $this->em->transactional(function(EntityManagerInterface $em) use ($executor, $fixtures, $append) {
-            if ($append === false) {
-                $executor->purge();
+        if($single_transaction) {
+            $executor = $this;
+            $this->em->transactional(function(EntityManager $em) use ($executor, $fixtures, $append) {
+                if ($append === false) {
+                    $executor->purge();
+                }
+                foreach ($fixtures as $fixture) {
+                    $executor->load($em, $fixture);
+                }
+            });
+        } else {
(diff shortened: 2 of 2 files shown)


lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
```diff
@@ -141,6 +141,7 @@ public function purge()
      *
      * @param array $fixtures Array of fixtures to execute.
      * @param boolean $append Whether to append the data fixtures or purge the database before loading.
+     * @param boolean $single_transaction Whether to use a single transaction when loading fixtures
      */
-    abstract public function execute(array $fixtures, $append = false);
+    abstract public function execute(array $fixtures, $append = false, $single_transaction = true);
```
KingCrunch: I know at least one project, that extends the executor, so this may lead to unexpected behaviour (BC). One can pass `false` as second argument, but the concrete executor class ignores it.

gnat42: If someone extends a class and overrides one of its methods, it would have to update the method signature because of the BC. I just re-submitted this because I was asked to rebase my previous PR.

skillberto:
Guys, whats going on with this?
I need this modification but it's still under merging since 2 years...
What's the wrong?

ctaverne:
Hi,

Why not merged for so long ?
It could be interesting for us to reduce memory usage on big imports.

stof:
@ctaverne because this is a BC break

ctaverne:
Ok, so maybe you should remove option:
      --multiple-transactions  Use one transaction per fixture file instead of a single transaction for all

in the command help

BafS:
Any new about this PR? It would be great to have it merged, if possible.

BafS (approved):

BafS:
Do we plan to merge it soon? If it's BC, what is the timeline for the next major? The PR is there since 7 years

Could we add the "v2.0.0" milestone tag?

greg0ire:
@BafS we don't plan to merge it since many things are wrong:
- the target branch is wrong as you pointed out yourself;
- there are no docs;
- there are no tests;
- there are conflicts.

Regarding the breaking change, I believe it could be worked around by getting the argument with `func_get_args()`.

> The PR is there since 7 years

So what? It has issues. Fix them if you want to see this merged.

derrabus:
Closing in favor of #388.

Issue #214, opened 2016-01-05
Deterministic order of data fixtures


sustmi:
When you have two data fixtures implementing `OrderedFixtureInterface` that have the same `order`, the order of their execution is non-deterministic.
That means that each time you run the loader the data fixtures could be loaded in different order.
This is a problem because when you have data fixtures that have to be loaded in certain order, but you forget to set the `order`, you sometimes get a correct result and sometimes a wrong one.

It happened to me that I tested certain scenario in development and it worked and then you deployed the same code to production and it did not work because the fixtures were loaded in different order.

The same applies to data fixtures implementing `DependentFixtureInterface` which are not dependent on each other.

I propose to implicitly order data fixtures alphabetically by name, so that when the order is not determined by `order` or dependencies it is still deterministic.

sustmi:
I think that the only non-deterministic input is the `get_declared_classes()` call in `Loader::loadFromIterator()`. I guess that ordering its result should be sufficient. The resulting data fixtures order will not be alphabetical as I suggested, but should be deterministic.

PR #215, opened 2016-01-05
Make the order of data fixtures loading deterministic

sustmi:
Fixes https://github.com/doctrine/data-fixtures/issues/214 .

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -350,6 +350,9 @@ class Loader
 
         $fixtures = array();
         $declared = get_declared_classes();
+        // Make the declared classes order deterministic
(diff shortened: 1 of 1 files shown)


KingCrunch:
There are already two interfaces [OrderedFixtureInterface](https://github.com/doctrine/data-fixtures#orderedfixtureinterface) and [DependentFixtureInterface](https://github.com/doctrine/data-fixtures#dependentfixtureinterface) that should solve your use-case, don't they?

sustmi:
Issue https://github.com/doctrine/data-fixtures/issues/214 is about situation when you **forget** to set all the order and dependencies. In such situation I would like the order to be always the same and not different on every run/machine (given probably by order of inclusion of PHP files).

mikeSimonson:
@sustmi Thanks

PR #222, opened 2016-05-05
Bring own implementation of TopologicalSorter to data-fixtures library

guilhermeblanco:
Fixes #212

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         "doctrine/common": "~2.2"
     },
     "require-dev": {
-        "doctrine/orm": "~2.4"
+        "doctrine/orm": "~2.4",
+        "phpunit/phpunit": "^4.0|^5.0"
     },
     "conflict": {
         "doctrine/orm": "< 2.4"
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": ">=5.3.2",
+        "php": "^5.5 || ^7.0",
         "doctrine/common": "~2.2"
     },
     "require-dev": {
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -101,9 +101,8 @@ class ORMPurger implements PurgerInterface
     public function purge()
     {
         $classes = array();
-        $metadatas = $this->em->getMetadataFactory()->getAllMetadata();
 
-        foreach ($metadatas as $metadata) {
+        foreach ($this->em->getMetadataFactory()->getAllMetadata() as $metadata) {
             if (! $metadata->isMappedSuperclass && ! (isset($metadata->isEmbeddedClass) && $metadata->isEmbeddedClass)) {
                 $classes[] = $metadata;
             }
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -157,35 +157,39 @@ class ORMPurger implements PurgerInterface
 
             // $class before its parents
             foreach ($class->parentClasses as $parentClass) {
-                $parentClass = $em->getClassMetadata($parentClass);
+                $parentClass     = $em->getClassMetadata($parentClass);
+                $parentClassName = $parentClass->getName();
 
-                if ( ! $sorter->hasNode($parentClass->name)) {
-                    $sorter->addNode($parentClass->name, $parentClass);
+                if ( ! $sorter->hasNode($parentClassName)) {
+                    $sorter->addNode($parentClassName, $parentClass);
                 }
 
-                $sorter->addDependency($class->name, $parentClass->name);
+                $sorter->addDependency($class->name, $parentClassName);
             }
 
             foreach ($class->associationMappings as $assoc) {
                 if ($assoc['isOwningSide']) {
-                    $targetClass = $em->getClassMetadata($assoc['targetEntity']);
+                    /* @var $targetClass ClassMetadata */
+                    $targetClass     = $em->getClassMetadata($assoc['targetEntity']);
+                    $targetClassName = $targetClass->getName();
 
-                    if ( ! $sorter->hasNode($targetClass->name)) {
-                        $sorter->addNode($targetClass->name, $targetClass);
+                    if ( ! $sorter->hasNode($targetClassName)) {
+                        $sorter->addNode($targetClassName, $targetClass);
                     }
 
                     // add dependency ($targetClass before $class)
-                    $sorter->addDependency($targetClass->name, $class->name);
+                    $sorter->addDependency($targetClassName, $class->name);
 
                     // parents of $targetClass before $class, too
                     foreach ($targetClass->parentClasses as $parentClass) {
-                        $parentClass = $em->getClassMetadata($parentClass);
+                        $parentClass     = $em->getClassMetadata($parentClass);
+                        $parentClassName = $parentClass->getName();
 
-                        if ( ! $sorter->hasNode($parentClass->name)) {
-                            $sorter->addNode($parentClass->name, $parentClass);
+                        if ( ! $sorter->hasNode($parentClassName)) {
+                            $sorter->addNode($parentClassName, $parentClass);
                         }
 
-                        $sorter->addDependency($parentClass->name, $class->name);
+                        $sorter->addDependency($parentClassName, $class->name);
                     }
                 }
             }
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -121,9 +121,9 @@ class ORMPurger implements PurgerInterface
             $class = $commitOrder[$i];
 
             if (
-                ($class->isInheritanceTypeSingleTable() && $class->name != $class->rootEntityName) ||
                 (isset($class->isEmbeddedClass) && $class->isEmbeddedClass) ||
-                $class->isMappedSuperclass
+                $class->isMappedSuperclass ||
+                ($class->isInheritanceTypeSingleTable() && $class->name !== $class->rootEntityName)
             ) {
                 continue;
             }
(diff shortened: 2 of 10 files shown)


tests/Doctrine/Tests/Mock/Node.php
```diff
@@ -0,0 +1,44 @@
...
+     * @var mixed
+     */
+    public $value;
+
+    /**
+     * Constructor.
+     *
+     * @param mixed $value
+     */
+    public function __construct($value = 0)
```
Ocramius: The default value is never used: kill it with fire

composer.json
```diff
@@ -13,7 +13,8 @@
         "doctrine/common": "~2.2"
     },
     "require-dev": {
-        "doctrine/orm": "~2.4"
+        "doctrine/orm": "~2.4",
+        "phpunit/phpunit": "~4.0"
```
Ocramius: Can you bump this to `^4.0|^5.0`?

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -145,48 +145,48 @@ public function purge()
...
-            $calc->addClass($class);
+            $calc->addNode($class->name, $class);
 
             // $class before its parents
             foreach ($class->parentClasses as $parentClass) {
                 $parentClass = $em->getClassMetadata($parentClass);
 
-                if ( ! $calc->hasClass($parentClass->name)) {
-                    $calc->addClass($parentClass);
+                if ( ! $calc->hasNode($parentClass->name)) {
```
bigfoot90: Undefined method ```TopologicalSorter::hasNode()```

guilhermeblanco: Saw that now... addressing

guilhermeblanco: Fixed

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -0,0 +1,198 @@
...
+ * TopologicalSorter is an ordering algorithm for directed graphs (DG) and/or
+ * directed acyclic graphs (DAG) by using a depth-first searching (DFS) to
+ * traverse the graph built in memory.
+ * This algorithm have a linear running time based on nodes (V) and dependency
+ * between the nodes (E), resulting in a computational complexity of O(V + E).
+ *
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
+ * @author Roman Borschel <roman@code-factory.org>
+ */
+class TopologicalSorter
```
Ocramius: As mentioned in chat - the fact that this is a Topological sorter is fine, but the fact that we are ordering fixture instances by topological order is unclear. @guilhermeblanco anything we can do for naming, either here or on the instance side to make that clear?

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * TopologicalSorter is an ordering algorithm for directed graphs (DG) and/or
+ * directed acyclic graphs (DAG) by using a depth-first searching (DFS) to
+ * traverse the graph built in memory.
+ * This algorithm have a linear running time based on nodes (V) and dependency
+ * between the nodes (E), resulting in a computational complexity of O(V + E).
+ *
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
+ * @author Roman Borschel <roman@code-factory.org>
```
Ocramius: This class should be marked as `@internal`. While it might seem useful, we don't want people to rely on it outside of `data-fixtures`.

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -0,0 +1,198 @@
...
+            switch ($childDefinition->state) {
+                case self::VISITED:
+                    continue;
+
+                case self::IN_PROGRESS:
+                    $message = 'Graph contains cyclic dependency. An example of this problem would be the following: '
+                        . 'Class C has class B as its dependency. Then, class B has class A has its dependency. '
+                        . 'Finally, class A has class C as its dependency.';
+
+                    throw new \RuntimeException($message);
```
Ocramius: Can we please throw a custom exception here (from the data-fixtures namespace, at least)?

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -0,0 +1,198 @@
...
+     * Adds a new node (vertex) to the graph, assigning its hash and value.
+     *
+     * @param string $hash
+     * @param object $node
+     *
+     * @return void
+     */
+    public function addNode($hash, $node)
+    {
+        $definition = new \stdClass();
```
Ocramius: As mentioned in chat, a tiny VO could make miracles to simple autocompletion and introspection here

tests/Doctrine/Tests/Common/DataFixtures/Sorter/TopologicalSorterTest.php
```diff
@@ -0,0 +1,134 @@
...
+
+namespace Doctrine\Test\DataFixtures\Sorter;
+
+use Doctrine\Common\DataFixtures\Sorter\TopologicalSorter;
+use Doctrine\Tests\Mock;
+
+/**
+ * TopologicalSorter tests.
+ *
+ * {@internal When writing tests here consider that a lot of graph
```
Ocramius: This is a note, not an `@internal`. @guilhermeblanco, `@internal` in phpdoc is meant to mark a symbol for internal use: it's not an actual comment for developers, rather than a warning for consumers.

tests/Doctrine/Tests/Mock/Node.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\Tests\Mock;
+
+/**
+ * Node.
```
Ocramius: Potato :P

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
guilhermeblanco: I don't think we can do further. Topological sorting is a known algorithm and a generic data structure.
We take a name/key/hash that is used as the vertex reference and the value can be anything you want. I used the fixture instance as value because it's easier and used its class name as the hash.

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
guilhermeblanco: The problem of using @internal is that it has different meaning for phpDocumentor.
This tag is used to prevent API generation exporting. We use that extensively in ORM, but while it's semantically correct to use it, the API generation tool considers it wrong.

I'll add it anyway... =)

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
guilhermeblanco: I don't think we need it. Fixture is supposed to be ran on its own and having a specific Exception brings no extra value. It's not that you ever want to capture only Fixture exceptions but not others.

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -0,0 +1,198 @@
...
+        $definition = $this->nodeList[$fromHash];
+
+        $definition->dependencyList = $dependencyList;
+    }
+
+    /**
+     * Return a valid order list of all current nodes.
+     * The desired topological sorting is the postorder of these searches.
+     *
+     * {@internal Highly performance-sensitive method.}
```
Ocramius: Should not be an `@internal` comment

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
Ocramius: Alright, seems reasonable, but we should set a standard for the next major versions (considering `Throwable` support)

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
Ocramius: Ok, can then the `$calc` on the instance side be addressed?

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
Ocramius: `@internal` causes most tooling to raise warnings on consumers, which is what is intended here.

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Sets the dependency list (edges) to the graph using their hashes.
+     *
+     * @param string $fromHash
+     * @param array  $dependencyList
+     *
+     * @return void
+     */
+    public function setDependencyList($fromHash, array $dependencyList)
```
Ocramius: Unused

fredw:
Do you have any prediction when this milestone will be released?
I ask this because I'm having the problems described in issue #212.

Ocramius:
Taking ownership of this branch for today.

Ocramius:
Tests fail due to PHP 5.6 requirements. Fixing in a separate PR in a sec.

Oliboy50:
So what are the compatible version of doctrine libs to avoid having the following issue? 
```
[Doctrine\Common\DataFixtures\Exception\CircularReferenceException]                                                                                                           
  Graph contains cyclic dependency. An example of this problem would be the following: Class C has class B as its dependency. Then, class B has class A has its dependency. Fi  
  nally, class A has class C as its dependency.
```

because I'm having this one with `data-fixtures` version 1.2.0 but I don't have this with `data-fixtures` version 1.1.1

PR #224, opened 2016-06-15
Added the option to exclude tables/views from purge

YaoOcelotl:


mikeSimonson:
@YaoOcelotl If you need to exclude tables or views, you need to use the filter option from configuration class of doctrine. If the entity manager that you pass to the data fixture is configured with the connection that exclude your tables you won't have any issue.

PR #225, opened 2016-06-16
Add ability to exclude tables from purge in ORM

YaoOcelotl:
Tests for  the  filter schema expression if present in the DBAL connection and checks is the table name has been explicitly excluded.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -141,18 +141,18 @@ class ORMPurger implements PurgerInterface
             $orderedTables[] = $this->getTableName($class, $platform);
         }
 
-        $connection = $this->em->getConnection();
-        $filterExpr = $connection->getConfiguration()->getFilterSchemaAssetsExpression();
-        $efe = empty($filterExpr);
-        foreach($orderedTables as $tbl) {
-            if(($efe||( !$efe && preg_match($filterExpr, $tbl) )) && array_search($tbl, $this->excluded)  === false){
-                if ($this->purgeMode === self::PURGE_MODE_DELETE) {
-                    $connection->executeUpdate('DELETE FROM ' . $tbl);
-                } else {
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -133,11 +142,15 @@ public function purge()
         }
 
         $connection = $this->em->getConnection();
+        $filterExpr = $connection->getConfiguration()->getFilterSchemaAssetsExpression();
+        $efe = empty($filterExpr);
```
Ocramius: What does `efe` mean?

mikeSimonson:
What I was trying to say was that because the filter schema option is available in the connection configuration, you won't see any of the tables that are filtered out by the connection.

Thus there is no point in adding that feature in all the app that use doctrine. It's already there.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
YaoOcelotl: Empty filter expression

Ocramius: Do not abbreviate then

YaoOcelotl:
The is filter expression regexp used during create schema operations not during query's.
 
See [setFilterSchemaAssetsExpression](http://www.doctrine-project.org/api/dbal/2.4/source-class-Doctrine.DBAL.Configuration.html#87-99)

YaoOcelotl:
The idea for filter expression regexp is that if tables/views should not be considered during schema creation they also should not be considered during schema purge.

As an example i use ~^(?!_view_)~ to avoid overrating view definitions.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
YaoOcelotl: Ok, i changed the variable name and added tests for exclusion using the filter expression or the explicit exclusion list.

mikeSimonson:
@YaoOcelotl Thanks

PR #226, opened 2016-06-19
Drop PHP 5.5 support, cleanup test suite according to PHPUnit 5.4 changes

Ocramius:


Ocramius:
BAM, gone.

mikeSimonson:
bam thanks

PR #229, opened 2016-06-20
Added support for self referencing classes in the TopologicalSorter

### LINKED ISSUE #228: TopologicalSorter breaks fixture load
The TopologicalSorter throws the error:

````
 Graph contains cyclic dependency. An example of this problem would be the following: Class C has class B as its dependency. Then, class B has class A has its dependency. Finally, class A has class C as its dependency. 
````

Everything was working before. As the error is not providing much information I will investigate further.

peterrehm:
A first solution to the issue #228. I don't know if this raises further issues.

- fixes #228

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
+++ b/lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
@@ -146,7 +146,7 @@ class TopologicalSorter
             $childDefinition = $this->nodeList[$dependency];
 
             // allow self referencing classes
-            if ($definition->value->getName() === $childDefinition->value->getName()) {
+            if ($definition === $childDefinition) {
                 continue;
             }
 
--- a/lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
+++ b/lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
@@ -145,6 +145,11 @@ class TopologicalSorter
 
             $childDefinition = $this->nodeList[$dependency];
 
+            // allow self referencing classes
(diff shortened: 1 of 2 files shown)


Ocramius:
Requires tests

peterrehm:
Just wanted to know if the general implementation makes sense. Then I can add a test.

peterrehm:
@Ocramius Is that test case enough?

Ocramius:
Seems good to me. @guilhermeblanco can you check this? I don't know if this follows your train of thought.

guilhermeblanco:
Reviewed and LGTM. Merged

NeilJ247:
Hi,

Not sure if this is the correct place to post if it isn't then let me know and I will raise as a new issue.  We are using Alice fixtures and still getting the same CircularReferenceException.  I was under the impression this was fixed in 1.2.1?  These are the versions for Alice fixtures and data-fixtures that we have been using:

        "hautelook/alice-bundle": "^1.3",
        "doctrine/data-fixtures": "^1.2",

Downgraded data-fixtures to 1.1 as a temporary fix within our project which doesn't give us the error.

Any help/feedback would be appreciated.

Thanks.

peterrehm:
Did you try with dev-master to test https://github.com/doctrine/data-fixtures/pull/240?

mikeSimonson:
I need to make a release. I will try to do that rapidly.

NeilJ247:
@peterrehm Hi sorry for the delayed answer.  Yes I have tried with dev-master and it works.  I thought it was fixed in the current version.

peterrehm:
@NeilJ247 There us a release peding by @mikeSimonson

NeilJ247:
@peterrehm thanks.

PR #233, opened 2016-06-22
Improved CircularReferenceException

peterrehm:


lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
```diff
@@ -156,9 +156,14 @@ private function visit(Vertex $definition)
...
                     throw new CircularReferenceException(
-                        'Graph contains cyclic dependency. An example of this problem would be the following: '
-                        . 'Class C has class B as its dependency. Then, class B has class A has its dependency. '
-                        . 'Finally, class A has class C as its dependency.'
+                        sprintf(
+                            'Graph contains cyclic dependency between the classes "%s" and'
+                            .' "%s". An example of this problem would be the following: '
+                            .'Class C has class B as its dependency. Then, class B has class A has its dependency. '
+                            .'Finally, class A has class C as its dependency.',
+                            $definition->value->getName(),
```
mgonyan: @peterrehm You will need to add a getName() method to the class Doctrine\Tests\Mock\Node in order to make the tests pass. However, do you know why we don't user a mock library rather than create a new specific class ?

peterrehm: No clue. Thanks for the feedback, will have a look at it on Monday.

peterrehm:
Should be good to merge now.

tests/Doctrine/Tests/Mock/Node.php
```diff
@@ -33,12 +33,26 @@ class Node
...
     public function __construct($value)
     {
         $this->value = $value;
+        $this->name = 'Node ' . (string) $value;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getName()
```
Ocramius: just use the public property?

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
Ocramius: `value` may be `mixed`. What is going on here?

peterrehm: Isn't it usually a ClassMetadata? What could be done is just show the proper exception if it is ClassMetadata otherwise show just the Exception as it used to be

Ocramius: `usually`

peterrehm:
Just updated.

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
peterrehm: What do you mean?

peterrehm:
@Ocramius WDYT now?

Ocramius:
@peterrehm you could modify the value signature to only accept `ClassMetadata` instances. Since the class is only internal, you can apply BC breaks to it as much as you want.

peterrehm:
@Ocramius just updated.

peterrehm:
@Ocramius Any comments?

Ocramius:
@guilhermeblanco can you check this? Is it ok for you to make the topological sorter ORM-specific?

guilhermeblanco:
Overall I have no strong opinion of enforcing `Vertex` to hold `ClassMetadata` instances, but I do recommend to switch from `ClassMetadataInfo` to `ClassMetadata`. The -Info one already does not exist on our develop branch of ORM.

peterrehm:
Updated.

lib/Doctrine/Common/DataFixtures/Sorter/TopologicalSorter.php
Ocramius: Note: this discussion was completed elsewhere - basically, we agreed that the `TopologicalSorter` can be `Doctrine\ORM\Mapping\ClassMetdata`-specific.

peterrehm: Yes, However I don't really get your comment. Everything is refactored to be `Doctrine\ORM\Mapping\ClassMetdata`-specific?

Ocramius: I just commented because it seemed like further work was needed, while all has been dealt with ;-)

peterrehm: Perfect then :)

Ocramius:
👍

Issue #230, opened 2016-06-21
TopologicalSorter breaks purge with foreign key constraints


peterrehm:
It breaks with SQL errors regarding Integrity constraint violations.

Related to #228.

dominikzogg:
@Ocramius thanks for the new version, we have the same problem on purge

```
Method setUpFixtures (Namespace\To\TestClass) has encountered an PDO Exception: An exception occurred while executing 'DELETE FROM entityname':
     [exec] 
     [exec] SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`databasename`.`some_relation_table`, CONSTRAINT `FK_.....` FOREIGN KEY (`parent_id`) REFERENCES `yetanothertable` (`id`))
```

Ocramius:
Description and error reports are not sufficient without an actual example or test-case.

Ocramius:
> @Ocramius the order to cleanup the tables does not or not complete consider the foreign_keys

Yes, that is understood.

 1. what are the defined fks?
 2. what are the defined tables/entities?
 3. what are the SQL operations performed
 4. in which order are they queued? (required to see how ordering logic needs to be adjusted)

mgonyan:
@Ocramius i think the error is  due to:
- The getCommitOrder function if it adds a node again it overrides its relations
- The sort method of TopologicalSorter gives us the list of tables already ordered, so they don't need to be ordered again. 
I am working on the test, so please have a look at the PR #231

dominikzogg:
@Ocramius https://github.com/doctrine/data-fixtures/pull/234, that is what i would expect, trying to delete a Role

mgonyan:
@dominikzogg why would you expect that output  (`users.role_id` column) when the function `getAssociationTables` returns all the joinTables to be truncate/delete their content as a first step in the purge process ?

dominikzogg:
@mgonyan As you can see, the other test, returns the same format. Deleteting a role is not possible in this case, without setting the column users.role_id to null or delete the users (see existing test classes).

Snowbaha:
Hi,

I have the same problem since I upgraded mysql to 5.5.49 (before 5.5.41) and same issue with mysql 5.7.*

I don't understand why a "minor" version can break all the fixture with the purge... :(

peterrehm:
[...]

PR #234, opened 2016-06-22
unidirectional-relation

dominikzogg:


mikeSimonson:
@dominikzogg What does that test try to show ?

dominikzogg:
@mikeSimonson Without setting null users.role_id or delete the row, roles arent deletable even if its a unidirectional relation

mikeSimonson:
@dominikzogg Can you add that in a comment on top of the test ?

dominikzogg:
@mikeSimonson done

mikeSimonson:
@dominikzogg Did you see that the test fails ?

stof:
What is the issue you have exactly ? I don't understand which issue you try to solve

dominikzogg:
@mikeSimonson as @stof mentioned, should i add a test or describe the whole scenario (which is not possible) the cleanup does not work. I my point of view, the code should be changed, that the test will run, cause at the moment unidirectional relations leads to trouble on data cleanup.

dominikzogg:
sorry not @stof mentioned it, it was @Ocramius https://github.com/doctrine/data-fixtures/issues/230#issuecomment-227435471

PR #238, opened 2016-06-30
Add a getFixture method

mgirouard:
I'm not sure if this is a good idea, but I've run into the case a few times where you need to get a fixture instance you didn't directly instantiate. Most often this happens when one fixture is automatically added by the loader as a dependency on another fixture.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -119,7 +119,10 @@ class Loader
     public function getFixture($className)
     {
         if (isset($this->fixtures[$className])) {
-            throw new \InvalidArgumentException(sprintf('"%s" is not a registered fixture'));
+            throw new \InvalidArgumentException(sprintf(
+                '"%s" is not a registered fixture',
+                $className
+            ));
         }
 
         return $this->fixtures[$className];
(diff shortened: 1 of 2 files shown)


mgirouard:
Any feedback on this? Happy to close if it's not useful!

Ocramius:
@mgirouard can you show where/how this would be used?

mgirouard:
We ran into a a problem where we didn't have full control over the `addFixture()` calls on a Loader. More specifically, we didn't have control over the order in which fixtures were added. Most of our fixtures implement `DependentFixtureInterface` so we couldn't rely on constructor. 

Instead, we have to do something similar to:

```php
foreach ($loader->getFixtures() as $fixture) {
    if ($fixture instanceof Fixtures\Foo) {
        $setting = $container->get('settings')['fooSetting'];
        $fixture->configureSetting($setting);
        break;
    }
}
```

Ideally we can just ask for the fixture by FQCN:

```php
$fixture = $loader->getFixture(Fixtures\Foo::class);
```

We've tried to restructure a bit, but this seems to be the simplest possible solution.

mikeSimonson:
@Ocramius Seems good to go to me ?

Ocramius:
Indeed, good to go.

mgirouard:
Thanks all. I'm looking forward to using this right away!

PR #231, opened 2016-06-21
Fixing foreign key constraints error

mgonyan:
this PR fixes https://github.com/doctrine/data-fixtures/issues/230

PR #239, opened 2016-07-15
Fix the sorting of metadata to be purged

stof:
The node for a class could have been added in the TopologicalSorter before the class is processed itself, in case it is the target of an association in a previous class.
In this case, the node should not be overwritten, as this would make it forget all previous dependencies.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -165,7 +165,9 @@ class ORMPurger implements PurgerInterface
         $sorter = new TopologicalSorter();
 
         foreach ($classes as $class) {
-            $sorter->addNode($class->name, $class);
+            if ( ! $sorter->hasNode($class->name)) {
+                $sorter->addNode($class->name, $class);
+            }
 
             // $class before its parents
             foreach ($class->parentClasses as $parentClass) {


stof:
So to be clear, the only case working properly currently when association are involved is when each class is processed before its dependencies, i.e. the input array is already sorted. An unsorted input would produce a broken output (which is quite unfortunate for a sorter)

Ocramius:
@stof can you write a test for it too?

stof:
Well, I don't understand how the ORMPurger is tested currently. It seems to be mocking the EM, but this bug will appear because of DB foreign keys.

Ocramius:
Ah, indeed, confused this with the sorter itself. You are just using the sorter here. Wondering if this rule should be encapsulated in `addNode()`?

stof:
yeah, the adder should probably check for it. Do you prefer for it to throw an exception if called a second time, or to silently keep the existing vertex ?

Ocramius:
Would probably keep the vertex, if it's the same...

mgonyan:
@stof, @Ocramius please have a look at https://github.com/doctrine/data-fixtures/pull/231

Tarsonius:
Is there any news on this?
Would love to have this merged..

stof:
Closing as #240 was merged and it also included this fix

PR #241, opened 2016-08-05
handle references for PHPCR ODM documents

takeit:
Doing:

``` php
$this->addReference('test', $document);
```

throws an exception:

```
[Symfony\Component\Debug\Exception\UndefinedMethodException]
  Attempted to call an undefined method named "getDocumentIdentifier" of class "Doctrine\ODM\PHPCR\UnitOfWork".
```

when using references with PHPCR ODM documents.

The PHPCR ODM UnitOfWork class doesn't have the method `getDocumentIdentifier`. There is `getDocumentId` instead. MongoDB ODM seems to use `getDocumentIdentifier` method in its UnitOfWork class but not in PHPCR.

Please, let me know if its enough.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
@@ -86,6 +86,11 @@ class ReferenceRepository
         if (method_exists($uow, 'getEntityIdentifier')) {
             return $uow->getEntityIdentifier($reference);
         }
+        
+        // Dealing with PHPCR ODM UnitOfWork
+        if (method_exists($uow, 'getDocumentId')) {
+            return $uow->getDocumentId($reference);
+        }
 
         // ODM UnitOfWork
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@takeit Is this the only thing that prevent you to use the data-fixture with the ODM ?

takeit:
@mikeSimonson I use it with PHPCR ODM in project I work on anyway. Even without the references that't not a blocker for me atm and it doesn't prevent me to use it.

mikeSimonson:
@takeit Can you add a test for it ?

kojiDev:
@mikeSimonson This issue seems obsolete:
https://github.com/doctrine/data-fixtures/blob/3a1e2c3c600e615a2dffe56d4ca0875cc5233e0a/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php#L68

alcaeus:
Thanks @kojiDev! This was indeed fixed in #255. Closing here.

PR #242, opened 2016-08-24
Make sure entities that are marked readOnly are excluded from purge

dhaarbrink:
In my case this entity is implemented as a view, DELETE or TRUNCATE won't work.
Whatever the case, readOnly implies no changing the data.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -112,7 +112,7 @@ class ORMPurger implements PurgerInterface
         $classes = array();
 
         foreach ($this->em->getMetadataFactory()->getAllMetadata() as $metadata) {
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@dhaarbrink Well the issue here is that apparently the readonly entities were not implemented to represent views. Don't know how the view should be represented nor how they should be treated by the data-fixture.

The easiest is probably to put those readonly entities (views) on a different connection/entity manager.
So that they are not touched by purge at all.

YaoOcelotl:
I use the filter schema expression from the DBAL connection.
See [Add ability to exclude tables from purge in ORM](https://github.com/doctrine/data-fixtures/pull/225)

```
# Doctrine Configuration
doctrine:
    dbal:
        schema_filter: ~^(?!view)~
```

dhaarbrink:
@YaoOcelotl thanks for the tip. Sadly it isn't released yet, but I did that myself now.
This was indeed what I needed, I will close this PR.

PR #250, opened 2017-01-19
ORMPurger: quote table names using reserved keywords

gigib82:
While DBAL supports auto quoting reserved keywords, ORMPurger doesn't actually make use of it.
This PR fixes this issue, test case included.

### DIFF
--- a/tests/Doctrine/Tests/Common/DataFixtures/TestEntity/ReservedKeyword.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/TestEntity/ReservedKeyword.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Doctrine\Tests\Common\DataFixtures\TestEntity;
+
+/**
+ * @Entity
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/Common/DataFixtures/TestEntity/ReservedKeyword.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Doctrine\Tests\Common\DataFixtures\TestEntity;
+
+/**
+ * @Entity
+ * @Table("SELECT")
```
Ocramius: Should be quoted here, as per https://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/limitations-and-known-issues.html#identifier-quoting-and-legacy-databases

stephen-lewis:
Looks to be fixed / implemented by #334 / #340 ?

PR #252, opened 2017-02-12
Allow to return empty arrays in getDependencies() method

javiereguiluz:
I've defined a base class for all my fixtures to simplify the application. This base class implements `DependentFixtureInterface` in case the fixture file needs to define its dependencies.

The problem is that some fixtures don't need this feature, so the base class returns an empty array in getDependencies() ... and that's not allowed.

Would you kindly consider merging this proposal? Thanks!

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -260,8 +260,8 @@ class Loader
                 
                 $this->validateDependencies($dependenciesClasses);
 
-                if (!is_array($dependenciesClasses) || empty($dependenciesClasses)) {
-                    throw new \InvalidArgumentException(sprintf('Method "%s" in class "%s" must return an array of classes which are dependencies for the fixture, and it must be NOT empty.', 'getDependencies', $fixtureClass));
+                if (!is_array($dependenciesClasses)) {
+                    throw new \InvalidArgumentException(sprintf('Method "%s" in class "%s" must return an array of classes which are dependencies for the fixture.', 'getDependencies', $fixtureClass));
                 }
 
                 if (in_array($fixtureClass, $dependenciesClasses)) {
--- a/tests/Doctrine/Tests/Common/DataFixtures/DependentFixtureTest.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/DependentFixtureTest.php
@@ -169,6 +169,19 @@ class DependentFixtureTest extends BaseTest
         $this->assertInstanceOf(__NAMESPACE__ . '\BaseParentFixture1', array_shift($orderedFixtures));
         $this->assertInstanceOf(__NAMESPACE__ . '\DependentFixture1', array_shift($orderedFixtures));
     }
+
+    public function test_dependentFixtureInterfaceAllowsToDefineEmptyDependencies()
+    {
+        $loader = new Loader();
+        $loader->addFixture(new NonEmptyDependenciesFixture);
+        $loader->addFixture(new EmptyDependenciesFixture);
+
+        $orderedFixtures = $loader->getFixtures();
+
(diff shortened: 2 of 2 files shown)


Ocramius:
Looking good, but needs tests

lcobucci:
@javiereguiluz @Ocramius isn't it weird to have a `DependentFixtureInterface` that doesn't have any dependency? Couldn't you implement the interface only on the fixtures that have dependencies?

I'd say this is more like a `won't fix`.

Ocramius:
If the dependencies are dynamic, then an empty set seems valid to me...

javiereguiluz:
@lcobucci I have defined this base class to abstract all the internal Doctine/Symfony details:

```php
abstract class Fixture extends AbstractFixture implements ContainerAwareInterface, DependentFixtureInterface
{
    use ContainerAwareTrait;

    public function getDependencies()
    {
        return array();
    }
}
```

This way, when a fixture file has dependencies, I define `getDependencies()`. Otherwise, I don't add that method ... and the base class returns the empty array.

Ocramius:
That's probably the worst counter-example to @lcobucci's remarks. The
reason why it failed so far on empty dependencies is so that people don't
foot-gun themselves while building dependency graphs, but you just made the
foot-gunning the base scenario 😲

javiereguiluz:
Then let's forget my comment and let's focus on Marco's remarks:

> If the dependencies are dynamic, then an empty set seems valid to me.

lcobucci:
Fixtures with static dependencies can already get very nasty be maintained, having a dynamic and potentially empty dependency list would increase the smell IMHO.

Despite that, it's an edge case that nobody missed for about 4 years...

Ocramius:
Given the lack of proper use-case scenario, I'll follow @lcobucci's advice and close this as `won't fix`. Indeed, dynamic fixtures would probably re-implement a fixture loader, rather than going through this approach.

weaverryan:
This is too bad - it's something that Javier and I worked on together. Our goal is to help make the fixtures library easier to use. I don't understand the foot-gunning comment - if you don't care / don't have dependencies, then what's wrong with pre-filling your deps as an empty array? Later, if/when you have deps, you would add them. This (Javier's code) is a classic example of adding an optional base to make an interface easier to use for most users. Is there some mistake the user can make with this that would cause real-world issues?

Thanks!

Ocramius:
Yes: spending hours debugging a dependant fixture not being loaded is a
real world scenario. The suggestion/example provided above is an
"abstract/base" fixture that doesn't really provide any value, unless I am
missing a scenario where it would make sense to have an empty set of
dependencies in a non-dynamic fixture loading process.

On 18 Feb 2017 2:58 p.m., "Ryan Weaver" <notifications@github.com> wrote:

This is too bad - it's something that Javier and I worked on together. Our
goal is to help make the fixtures library easier to use. I don't understand
the foot-gunning comment - if you don't care / don't have dependencies,
then what's wrong with pre-filling your deps as an empty array? Later,
if/when you have deps, you would add them. This (Javier's code) is a
classic example of adding an optional base to make an interface easier to
use for most users. Is there some mistake the user can make with this that
would cause real-world issues?

Thanks!

—
You are receiving this because you were assigned.

Reply to this email directly, view it on GitHub
<https://github.com/doctrine/data-fixtures/pull/252#issuecomment-280847240>,
or mute the thread

weaverryan:
Yea, you're right it adds limited value - it just saves you from needing to add the interface (just override the method). So, fair point :).

Hours of debugging? That feels like a stretch - fixtures are a pretty small problem :). But no worries - this was a really minor change, so it's cool either way. We *are* hoping to propose other DX changes to other parts of Doctrine - as long as you'll hear us out on those, I'm  very happy! Doctrine rocks... but it's hard for a lot of people!

Cheers!

PR #253, opened 2017-02-23
Fix syntax highlighting on README

fefas:
There is a _Code Block_ with no syntax highlighting and I noted that it was out of the this README's standard.

:)


Ocramius:
LGTM :+1:

PR #256, opened 2017-05-28
Use PSR-4 instead of PSR-0

fefas:
> PSR-0 is deprecated.

I moved every file to let the directory tree clean, but I can redo the commit and keep the directories unchanged.


Ocramius:
Doctrine project has voted "no" on PSR-4 due to its confusing directory structure mappings (when we were still in the FIG).

Closing here: patch can be accepted if re-opened with only changes to `composer.json` (no moved files).

Ocramius:
@fefas just open a PR without the moved files 👍

PR #257, opened 2017-05-28
Use PSR-4 instead of PSR-0

fefas:
PSR-0 is deprecated.

> Replace PR #256.


Ocramius:
LGTM!

Thanks @fefas!

PR #270, opened 2017-09-07
Add return type to Ordered Fixture Interface

radekhubner:


Majkl578:
Hello.
Adding a return type in an interface is BC break and thus can't go to 1.x. Consider re-targeting this PR to 2.0 branch and also adding note to the UPGRADE document.
Travis failure is unrelated, caused by https://github.com/mongodb/mongo-php-driver/issues/649.

PR #272, opened 2017-10-30
Disable foreign key checks on purge for all MySQL versions

### LINKED ISSUE #113: Support truncating tables with foreign keys

robertfausk:
... when purge mode is truncate

* add test for change of foreign key check when using MySQL
* add test for not set foreign key check when using SQLite

---

As statet in https://github.com/doctrine/dbal/pull/2620 doctrine/dbal is not responsible for this ("third party code has to built its own implementation around this because it is highly use-case specific").
So I give it a try to finish this long term never ending story. 

Closes (stale?) PR #127 and fixes #113.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\Common\DataFixtures\Purger;
 
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\AbstractMySQLDriver;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Common\DataFixtures\Sorter\TopologicalSorter;
 use Doctrine\ORM\Mapping\ClassMetadata;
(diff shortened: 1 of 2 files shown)


robertfausk:
Määähhh. Build currently fails because of pecl installation of mongodb. Also see #271.

At least nightly passed :hushed:

theofidry:
@ocramius I know we already talked about that and at that time you were not in favour of that choice. 

Maybe you changed your opinion since then, but I still think this should be done here. We are in the context of fixtures were truncating/deleting some tables don't make sense for the application.

We can always argue that the user can do that on his side (and that's what they end up doing), but it's silly to push that extra step to the user when 90% if not more needs it and the other 10% probably don't care.

On the same token, a recurrent request is to not truncate read-only tables. I have more mixed feeling about this but maybe this is actually a good idea and should be handled by the purger as well.

jopais:
I think this is an important change. Can you look at it again?

ilya-realforce:
Any update on this? I think it's straight-up `ORMPurger`'s responsibility to purge DB even if there're foreign keys.

PR #274, opened 2017-11-27
Symfony 4 support: Hook point to control creating fixtures objects

weaverryan:
Hi guys!

This small patch is needed to support Symfony 4. Specifically, in doctrine/DoctrineFixturesBundle#209, we re-work fixtures to make the classes proper services, which is an easy patch.

However, in this parent class, there are 2 spots where the fixture objects are *instantiated*. That's a problem: the *container* should instantiate them. With this patch, in the bundle, we will be able to throw an exception in `createFixture()` if the object is not already available as a service.

Thanks!


alcaeus:
Thanks @weaverryan!

PR #275, opened 2017-11-27
Use assertEmpty

gabriel-caruso:
Small one: I've refactored a test with [`assertEmpty`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEmpty) method.


alcaeus:
Build failure is unrelated, I'll fix that separately.

PR #278, opened 2017-12-10
Refactoring tests

carusogabriel:
I've refactored some tests, like #275, this time using:
- `assertInstanceOf` instead of `instanceof` operator;
- `assertCount` instead of `count` function;
- `assertGreaterThan` for mathematical comparisons;


mikeSimonson:
@carusogabriel Thanks

PR #280, opened 2017-12-14
Added access modifier to DependentFixtureInterface

joshlopes:
Implement the access modifier making it PSR-2 compliant.

Just noticed that using some IDE functionality to implement this interface copy/paste the code from it and later marks it as an "error". 
Because doctrine is using PSR-2, i think it makes sense making this one compliant too.


alcaeus:
Thanks @joshlopes

PR #284, opened 2018-02-13
clean up phpcr executor

dbu:
- use the DocumentManagerInterface 
- use php 7 static typing
- remove unnecessary phpdoc blocks

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Executor/PHPCRExecutor.php
+++ b/lib/Doctrine/Common/DataFixtures/Executor/PHPCRExecutor.php
@@ -2,7 +2,7 @@
 
 namespace Doctrine\Common\DataFixtures\Executor;
 
-use Doctrine\ODM\PHPCR\DocumentManager;
+use Doctrine\ODM\PHPCR\DocumentManagerInterface;
 use Doctrine\Common\DataFixtures\Purger\PHPCRPurger;
 
 /**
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Common/DataFixtures/Executor/PHPCRExecutor.php
```diff
@@ -14,26 +14,26 @@
...
-        parent::__construct($dm);
     }
 
-    /**
-     * Retrieve the DocumentManager instance this executor instance is using.
-     *
-     * @return \Doctrine\ODM\PHPCR\DocumentManager
-     */
-    public function getObjectManager()
+    public function getObjectManager(): DocumentManagerInterface
```
alcaeus: This is a a BC break: we still have `1.3.x` as branch-alias for master, so we shouldn't break BC (yet).

lib/Doctrine/Common/DataFixtures/Purger/PHPCRPurger.php
```diff
@@ -14,37 +13,21 @@
...
         $this->dm = $dm;
     }
 
-    /**
-     * Set the DocumentManager instance this purger instance should use.
-     *
-     * @param DocumentManager $dm
-     */
-    public function setDocumentManager(DocumentManager $dm)
+    public function setDocumentManager(DocumentManagerInterface $dm)
```
alcaeus: This is also a BC break.

lib/Doctrine/Common/DataFixtures/Purger/PHPCRPurger.php
```diff
@@ -14,37 +13,21 @@
...
         $this->dm = $dm;
     }
 
-    /**
-     * Retrieve the DocumentManager instance this purger instance is using.
-     *
-     * @return \Doctrine\ODM\PHPCR\DocumentManager
-     */
-    public function getObjectManager()
+    public function getObjectManager(): DocumentManagerInterface
```
alcaeus: One more BC break.

dbu:
is it intended use of this repository to extend the executor? there is only a BC break if somebody overwrites that method. guess we should have made those classes final to be sure. 

unfortunately, i deleted my fork some time ago. do you have any idea how i can edit this now? can i somehow download a diff to recreate what is in here?

alcaeus:
You can add `fetch = +refs/pull/*/head:refs/remotes/origin/pr/*` to the remote in `.git/config`, then checkout a new branch based on this PR.

alcaeus:
To follow up on the other part of your question, yes, we should have made a lot of classes final, but that's how it is. For now, they are part of the public API, so we can neither make them final nor can we change any method signatures (except for **some** changes in the constructor signature since they don't cause signature mismatches).

dbu:
thanks a lot for the git instructions, was not aware thats possible, pretty cool! continuing in #289

PR #287, opened 2018-03-14
Fix issue referencing a non string object ID

xaben:
When using Value Object as IDs in Doctrine, the ProxyReferenceRepository fails to unserialize the reference repository. This is caused by the fact that `json_encode` and `json_decode` does not know how to persist information about these objects (as it expects to have a plain data structure with string for IDs).

This PR solves it by using `serialize` and `unserialize` instead.


Ocramius:
Requires test additions

xaben:
@Ocramius Are the tests like this ok? 

I tried adding a separate entity and a separate test but the insert into DB always failed, perhaps I missed some setup procedure somewhere.

PR #289, opened 2018-03-19
clean up phpcr executor

dbu:
replace #284, without BC breaks.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/PHPCRPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/PHPCRPurger.php
@@ -2,9 +2,9 @@
 
 namespace Doctrine\Common\DataFixtures\Purger;
 
-use PHPCR\Util\NodeHelper;
-
 use Doctrine\ODM\PHPCR\DocumentManager;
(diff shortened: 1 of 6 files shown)


lib/Doctrine/Common/DataFixtures/Purger/PHPCRPurger.php
```diff
@@ -14,36 +14,20 @@
 class PHPCRPurger implements PurgerInterface
 {
     /**
-     * DocumentManager instance used for persistence.
-     * @var DocumentManager
+     * @var DocumentManagerInterface
```
Ocramius: @dbu in which version was this interface introduced? Should we add a `conflict` clause to `composer.json`?

dbu: it was added in 2013, was not aware that we had releases before that time :-D

its version 1.3.0, will add a conflicts

dbu:
but the test failure came from .travis.yml not installing phpcr-odm at all :-O

there is no minimal version build

dbu:
ok, now the build is green

Ocramius:
Thanks @dbu!

PR #291, opened 2018-03-29
fix(loader): don't create fixture dependency if already added

tomtomau:
I outlined this below on Slack:

Symfony4, DoctrineFixturesBundle 3.4, LiipFunctionalTestingBundle.
I have a Fixture A which depends on Fixture B.
Fixture B has constructor dependency injection going on, so the LiipFunctionalTesting bundle loads that from the container [(see here)](https://github.com/liip/LiipFunctionalTestBundle/blob/master/Test/WebTestCase.php#L713) and calls ->addFixture on the loader.

Tracing it through, I can see here that we're doing a plain ol instantiation of the class, even if the fixture is already loaded - [Loader.php#L139](https://github.com/doctrine/data-fixtures/blob/master/lib/Doctrine/Common/DataFixtures/Loader.php#L139)

We can also definitely fix this in the LiipFunctionalTestingBundle by creating a custom Loader and overriding the createFixture method to use the Container to resolve it instead, but it also seems logical to me that this package doesn't try to create another fixture if it has already been added.

Let me know your thoughts :)

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -135,7 +135,7 @@ class Loader
             } elseif ($fixture instanceof DependentFixtureInterface) {
                 $this->orderFixturesByDependencies = true;
                 foreach($fixture->getDependencies() as $class) {
-                    if (class_exists($class)) {
+                    if (!isset($this->fixtures[$fixtureClass]) && class_exists($class)) {
                         $this->addFixture($this->createFixture($class));
                     }
                 }
--- a/tests/Doctrine/Tests/Common/DataFixtures/LoaderTest.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/LoaderTest.php
@@ -2,9 +2,11 @@
 
 namespace Doctrine\Tests\Common\DataFixtures;
 
+use Doctrine\Common\DataFixtures\DependentFixtureInterface;
 use Doctrine\Common\DataFixtures\FixtureInterface;
 use Doctrine\Common\DataFixtures\Loader;
 use Doctrine\Common\DataFixtures\SharedFixtureInterface;
+use Doctrine\Common\Persistence\ObjectManager;
 use TestFixtures\MyFixture1;
 use TestFixtures\NotAFixture;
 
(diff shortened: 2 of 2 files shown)


jwage:
Right now if you register the same fixture class twice, will it try to execute it twice?

tomtomau:
@jwage not exactly.

If you try to call `->addFixture` on the same fixture twice, it will bail early on line 123.

The specific case I'm trying to catch here is if you have already loaded the dependency, we should not try to create the dependency again and rely on the ->addFixture method checking if we've already got it added.

alcaeus:
> DoctrineFixturesBundle 3.4

Wait, what? The newest release of `doctrine/doctrine-fixtures-bundle` is 3.0.2, there's no 3.4 release (yet). Taking a look at LiipFunctionalTestBundle, it seems the 2.x release supports this version of the fixtures bundle, so I'll base my analysis on that.

You're analysis is correct in the sense that `createFixture` does a good ole' instantiation of a fixture class (see [Loader:191..194](https://github.com/doctrine/data-fixtures/blob/3a1e2c3c600e615a2dffe56d4ca0875cc5233e0a/lib/Doctrine/Common/DataFixtures/Loader.php#L191..L194)), but the loader included in the fixtures bundle overrides this behavior, with `createFixture` returning a previously loaded fixture (see [SymfonyFixturesLoader:59..73](https://github.com/doctrine/DoctrineFixturesBundle/blob/b9c1c7f2afb4dfc3bcae2172dff93f1bfa5e94d3/Loader/SymfonyFixturesLoader.php#L59..L73)). The problem seems to be the `WebTestCase` class included in the new version of the FunctionalTestBundle: it either instantiates the `ContainerAwareLoader` from the Doctrine bridge or the `Loader` included in `doctrine/data-fixtures`, but never the new `SymfonyFixturesLoader` from the new bundle version (see [WebTestCase:684..691](https://github.com/liip/LiipFunctionalTestBundle/blob/master/Test/WebTestCase.php#L684..L691).

This library wasn't really intended to handle fixtures with constructor dependencies (as evidenced by the `createFixture` method instantiating the class) with multiple instances of the same fixture class loaded if necessary. I wouldn't change this behavior for now, but this is something we might revisit for version 2.0 of the library to better support the behavior needed by the Symfony bundle and other libraries. For now the fix would be to use the new `SymfonyFixturesLoader` class and always pass an instance of the fixture class when adding it.

PR #292, opened 2018-04-05
Quote tables properly when using delete mode

bendavies:
This changes `getTableName` to use the same method for quoting the table name as `$platform->getTruncateTableSQL(...)`.

This fixes the case where you have a table name which is a reserved word.


versgui:
A merge would be welcome.
For example, an error is triggered by Postgresql on a table called "group".

stephen-lewis:
I've created an alternative PR #334 that fixes this issue and includes tests.

PR #298, opened 2018-10-11
Add GroupsInterface to allow Fixtures to be loaded in groups

noemiquezada:


Ocramius:
@noemiquezada what was the original intent of this patch? Was it supposed to be closed?

PR #314, opened 2019-09-04
Use single query truncate when possible

goetas:
Hi, 
this PR allows to truncate multiple tables with a single query when supported (postgres as example).

I'm pretty sure that is not really elegant, but it allowed us to have a test suite running in ~20s instead of 6m. I'm open for suggestions on how to improve it.

_It would be nice to have `getTruncateTableSQL` allowing an array as table name..._

Did not write tests yet as I'm not sure that the current version is good and that will be considered at all for merging.

Feedback?

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Common\DataFixtures\Purger;
 
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Common\DataFixtures\Sorter\TopologicalSorter;
 use Doctrine\ORM\Mapping\ClassMetadata;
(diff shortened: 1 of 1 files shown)


goetas:
A first idea could be to add `isMultiTableTrucateSupported()` and relative methods in the DBAL package.

alcaeus:
This should definitely be solved in DBAL. Since that will take time (and may even require a BC break to have `getTruncateTableSQL` accept an array), it may be worth working around this restriction here for now. Other opinions @doctrine/doctrinecore?

ostrolucky:
I share same opinion

goetas:
>and may even require a BC break to have getTruncateTableSQL accept an array

Would be considered a BC break adding to the `AbstractPlatform` class a new method `getTruncateMultiTableSQL` ?

alcaeus:
> Would be considered a BC break adding to the `AbstractPlatform` class a new method `getTruncateMultiTableSQL` ?

As long as the method isn't `abstract`, I don't see a problem with that.

goetas:
Will start with a PR there then!

greg0ire:
I'm closing, but tell me if I need to reopen this.

PR #318, opened 2019-09-04
Fix: No need to update composer itself

localheinz:
This PR

* [x] stops updating `composer` itself on Travis (again)


alcaeus:
Thanks @localheinz!

localheinz:
Thank you, @alcaeus!

PR #331, opened 2019-12-13
Doctrine\Common\Persistence\ObjectManager is deprecated

pwnyprod:
... but you cant use the new Doctrine\Persistence\ObjectManager because of TypeHinting.

Found that on the FixtureBundle 3.3 in Symfony 5.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/FixtureInterface.php
+++ b/lib/Doctrine/Common/DataFixtures/FixtureInterface.php
@@ -4,7 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Common\DataFixtures;
 
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Common\Persistence\ObjectManager as DeprecatedManager;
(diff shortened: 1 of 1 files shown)


alcaeus:
There will be new releases for all affected projects to account for the deprecations. Thanks for bringing it up 👍

lib/Doctrine/Common/DataFixtures/FixtureInterface.php
```diff
@@ -4,7 +4,8 @@
 
 namespace Doctrine\Common\DataFixtures;
 
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Common\Persistence\ObjectManager as DeprecatedManager;
+use Doctrine\Persistence\ObjectManager;
```
greg0ire: These are not 2 different classes, one is actually an alias to the other one, so there is no issue with using the new name in the signature because it's the same thing

greg0ire:
Fixed in #333

PR #332, opened 2019-12-28
Add correct return type to DependentFixtureInterface::getDependencies

ruudk:
This makes it easier for PHPStan to understand what's going on.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/DependentFixtureInterface.php
+++ b/lib/Doctrine/Common/DataFixtures/DependentFixtureInterface.php
@@ -13,7 +13,7 @@ interface DependentFixtureInterface
      * This method must return an array of fixtures classes
      * on which the implementing class depends on
      *
-     * @return array
+     * @return class-string[]
      */
     public function getDependencies();
 }


alcaeus:
How does PhpStorm handle the class-string type? I’m worried that it may not understand what’s going on...

ruudk:
I checked. It just sees it as array of mixed, and that was already the case :)

Screenshot 2019-12-28 at 11 48 42

ruudk:
I'm sure PHPStorm will soon support all the new PHPStan/Psalm notations.

alcaeus:
Great. I’ll take care of this after the holidays, as I need to check the branches and do some more work here for the persistence deprecations.

I believe this should be fixed in a patch release, would you agree?

ruudk:
Thanks! I agree.

greg0ire:
@ruudk this means you shouldn't target master

ruudk:
@greg0ire Sorry, didn't get the hint. Changed the base branch now.

greg0ire:
Triggering a new build, since #333 fixed it

lib/Doctrine/Common/DataFixtures/DependentFixtureInterface.php
```diff
@@ -13,7 +13,7 @@ interface DependentFixtureInterface
      * This method must return an array of fixtures classes
      * on which the implementing class depends on
      *
-     * @return array
+     * @return class-string[]
```
garak: I think this should be `@return FixtureInterface[]`

ruudk: I thought so too, but it turns out it expects an array of FQCN class names to do the sorting. Not the actual instances.

ruudk: https://github.com/doctrine/data-fixtures/blob/608a35a3b5bcc4214d116603095f8b0c51091592/lib/Doctrine/Common/DataFixtures/Loader.php#L170-L176

garak: So, it should be `@return \Doctrine\Common\DataFixtures\DependentFixtureInterface[]|\Doctrine\Common\DataFixtures\FixtureInterface[]`

garak: That's a bit of mess, I really don't get why other interfaces are not extending `FixtureInterface`... :disappointed:

ruudk: It should not return instances of that type, it should return FQCN (strings) of fixtures to load first.

garak: ah, you're right. My bad

ruudk:
@greg0ire Can we merge it? :)

greg0ire:
I'm not sure: 1.3 is still maintained, isn't it affected by this too?

ruudk:
@greg0ire Changed base branch and rebased it :shipit:

greg0ire:
Aaaaand the test fail… please cherry-pick b7532e4, that should fix it.

ruudk:
Sorry, but that's not part of this PR. Clearly 1.3.x- branch is broken. I tried to cherry pick it but that forces me to fix a merge conflict. This such a small change, that I already spend too much time one.

greg0ire:
Ok don't worry, I'm going to take care of fixing the branch, just leave this open.

greg0ire:
Ok so apparently it is no longer maintained, so targeting 1.4.x was the right thing to do, sorry.

greg0ire:
Thanks @ruudk !

vitek-rostislav:
Hi, this made my phpstan in version `0.11.6` fail on
```
invalid type Doctrine\Common\DataFixtures\class
```
:disappointed: 

Luckily, upgrading to `^0.12` version helped.

PR #333, opened 2019-12-30
Address deprecations from persistence

greg0ire:
A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See doctrine/persistence#71

This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in type declarations of non-constructor methods,
so that signature compatibility is recognized by old versions of php.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf


greg0ire:
Seems like this has to do with https://github.com/doctrine/orm/pull/7079

composer.json
```diff
@@ -25,7 +26,7 @@
         "doctrine/coding-standard": "^6.0",
         "doctrine/dbal": "^2.5.4",
         "doctrine/mongodb-odm": "^1.3.0",
-        "doctrine/orm": "^2.5.4",
+        "doctrine/orm": "^2.7.0",
```
ostrolucky: Why are we bumping this?

greg0ire: It's in order to make sure we get the changes from this PR: https://github.com/doctrine/orm/pull/7079 (`getJoinTableName` is called inside `getAssociationTables`)

ostrolucky: Why is that relevant in this PR?

greg0ire: Because the tests will not pass otherwise. Do you want me to extract this commit in a separate PR?

ostrolucky: Nah it's allright it can be here, it's small

lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
```diff
@@ -139,3 +140,5 @@ public function purge()
      */
     abstract public function execute(array $fixtures, $append = false);
 }
+
+interface_exists(ObjectManager::class);
```
shadowhand: Why is this line included? It appears a couple of other places too, and I don't see why it is necessary.

greg0ire: It really isn't. If you want to know, follow the link in the PR body / the commit message.

PR #334, opened 2020-01-02
Quote table names that also are reserved keywords

stephen-lewis:
Technically a duplicate of PR #292 but with tests included.

Essentially, purging the database in DELETE mode prior to fixture load fails if any fixture entities use tables names with reserved words (e.g. 'User' in Postgres, or in the test case 'Group' in sqlite). The change makes ORMPurger->getTableName() use the same method as AbstractPlatform::getTruncateTableSQL() uses to get the table name (quoted as necessary).

Test case uses sqlite reserved word 'Group' as an example and checks returned table name is quoted correctly.

Closes #292

### DIFF
--- a/tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerTest.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerTest.php
@@ -15,6 +15,7 @@ class ORMPurgerTest extends BaseTest
     public const TEST_ENTITY_USER             = TestEntity\User::class;
     public const TEST_ENTITY_USER_WITH_SCHEMA = TestEntity\UserWithSchema::class;
     public const TEST_ENTITY_QUOTED           = TestEntity\Quoted::class;
+    public const TEST_ENTITY_GROUP            = TestEntity\Group::class;
 
     public function testGetAssociationTables()
     {
@@ -53,5 +54,19 @@ class ORMPurgerTest extends BaseTest
         $method->setAccessible(true);
         $tableName = $method->invokeArgs($purger, [$metadata, $platform]);
         $this->assertStringStartsWith('test_schema', $tableName);
+    }
+
+    public function testGetTableNameQuoted() : void
+    {
+        $em       = $this->getMockAnnotationReaderEntityManager();
+        $metadata = $em->getClassMetadata(self::TEST_ENTITY_GROUP);
+        $platform = $em->getConnection()->getDatabasePlatform();
+        $purger   = new ORMPurger($em);
+        $class    = new ReflectionClass(ORMPurger::class);
+        $method   = $class->getMethod('getTableName');
+        $method->setAccessible(true);
+        $tableName = $method->invokeArgs($purger, [$metadata, $platform]);
+        $this->assertStringStartsWith('"', $tableName);
+        $this->assertStringEndsWith('"', $tableName);
     }
 }
--- a/tests/Doctrine/Tests/Common/DataFixtures/TestEntity/Group.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/TestEntity/Group.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Common\DataFixtures\TestEntity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity()
+ */
+class Group
+{
+    /**
+     * @ORM\Column(type="integer")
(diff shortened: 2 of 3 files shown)


greg0ire:
The tests fail, but https://github.com/doctrine/data-fixtures/pull/333 fixes the build

tests/Doctrine/Tests/Common/DataFixtures/TestEntity/Group.php
```diff
@@ -0,0 +1,49 @@
...
+    public function setCode($code)
+    {
+        $this->code = $code;
+    }
+
+    public function getCode() : ?string
+    {
+        return $this->code;
+    }
+}
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

tests/Doctrine/Tests/Common/DataFixtures/TestEntity/Group.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $this->id = $id;
+    }
+
+    public function getId() : ?int
+    {
+        return $this->id;
+    }
+
+    public function setCode($code)
```
greg0ire:
```suggestion
    public function setCode($code) : void
```

tests/Doctrine/Tests/Common/DataFixtures/TestEntity/Group.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @ORM\Column(length=32)
+     * @ORM\Id
+     *
+     * @var string|null
+     */
+    private $code;
+
+    public function setId($id)
```
greg0ire:
```suggestion
    public function setId($id) : void
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerTest.php
```diff
@@ -54,4 +55,18 @@ public function testTableNameWithSchema()
         $tableName = $method->invokeArgs($purger, [$metadata, $platform]);
         $this->assertStringStartsWith('test_schema', $tableName);
     }
+
+    public function testGetTableNameQuoted()
```
greg0ire:
```suggestion
    public function testGetTableNameQuoted() : void
```

greg0ire:
#333 is not merged up yet, hence why the build is not fixed. Are you sure you should target master though? Is there a BC-break in your PR?

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerTest.php
stephen-lewis: I agree in principle with this, but the other methods on the class don't have return types - prefer consistency in this PR and then update all test class methods in another to include return types?

greg0ire: > update all test class methods in another to include return types

That will mean a bigger PR than it should. Consistency is great, but consistency with deprecated things isn't.

stephen-lewis: Good argument - changed

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerTest.php
```diff
@@ -54,4 +55,18 @@ public function testTableNameWithSchema()
...
+    {
+        $em       = $this->getMockAnnotationReaderEntityManager();
+        $metadata = $em->getClassMetadata(self::TEST_ENTITY_GROUP);
+        $platform = $em->getConnection()->getDatabasePlatform();
+        $purger   = new ORMPurger($em);
+        $class    = new ReflectionClass(ORMPurger::class);
+        $method   = $class->getMethod('getTableName');
+        $method->setAccessible(true);
+        $tableName = $method->invokeArgs($purger, [$metadata, $platform]);
+        $this->assertStringStartsWith('"', $tableName);
```
SenseException: Why not asserting the full table name?

stephen-lewis: I think I was focused on ensuring the quotes were there without having to worry about the case of the table name itself. As the table name isn't given in the entity, the function needs to work regardless of whether the table name is 'Group', or 'GROUP', or 'Group', etc. and this will prove that.

That said, the test is obviously run in a minimal config environment so the test entity's table name should default to 'Group'.

PR #335, opened 2020-01-09
Expect a table name that uses schema emulation

greg0ire:
We use SQLite in tests, and that platform does not support schemas.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "doctrine/common": "~2.2"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


greg0ire:
This is a backport of something I merged in 1.4.x, I didn't realise this branch was maintained but the metadata file in master says so.

alcaeus:
This branch is no longer maintained, looks like the metadata is wrong again 1.4.0 is the lowest supported branch.

PR #346, opened 2020-03-12
Fix return type in DependentFixtureInterface

groovytron:
This fixes the return type in the DependentFixtureInterface's `getDependencies()` method. `class-string[]` is not a valid type and makes static code analyzers like [PHPStan](https://github.com/phpstan/phpstan) to fail due to this invalid type.

I hope this fix helps.

Thanks for your feedback.


greg0ire:
The scrutinizer build is failing because it's using 7.4 code on a non-7.4 platform: https://github.com/Ocramius/PackageVersions/blob/master/src/PackageVersions/Installer.php#L34

SenseException:
@greg0ire In Scrutinizer itself, Settings -> Configuration, but if a `.scrutinizer.yml` file exists in the repository, this will be used instead. We should think about versioning of Scrutinizer files in repos.

I'm not sure if it is

``` yaml
build:
    environment:
        php:
            version: 7.4
```

or 


``` yaml
build:
        nodes:
            analysis:
                environment:
                    php:
                        version: 7.4
```

that influences the build. I have to look in the docs for that.

groovytron:
Is there something I can do to fix it in my pull request?

greg0ire:
I don't think you can. @SenseException , can you please fix this?

ruudk:
`class-string` _is_` supported by PHPStan, see https://github.com/phpstan/phpstan/commit/77eb14115b425213e38c91cf0b105fbadcb44901

groovytron:
> `class-string` _is_ supported by PHPStan, see [phpstan/phpstan@77eb141](https://github.com/phpstan/phpstan/commit/77eb14115b425213e38c91cf0b105fbadcb44901)

You are right. Updating PHPStan seems to have fixed the issue. Do you want to close this merge request then?

alcaeus:
Works as expected - let's hope this is the last time we're dealing with this ;)

PR #347, opened 2020-04-05
Allow Custom ORMPurger

kissifrot:
https://github.com/doctrine/DoctrineFixturesBundle/pull/307 on Bundle-side allowed to create custom Purger, however an instance of ORMPurger is needed.
This PR allows a new ORMPurgerInterface to be used instead.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurgerInterface.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurgerInterface.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Common\DataFixtures\Purger;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * ORMPurgerInterface
+ */
+interface ORMPurgerInterface
+{
+    /**
+     * Set the EntityManagerInterface instance this purger instance should use.
+     *
+     * @return void
+     */
+    public function setEntityManager(EntityManagerInterface $em);
+}
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurgerInterface.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurgerInterface.php
@@ -9,7 +9,7 @@ use Doctrine\ORM\EntityManagerInterface;
 /**
  * ORMPurgerInterface
  */
-interface ORMPurgerInterface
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Common/DataFixtures/Purger/ORMPurgerInterface.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Common\DataFixtures\Purger;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+/**
+ * ORMPurgerInterface
+ */
+interface ORMPurgerInterface
```
alcaeus, line 12: I believe this should extend `PurgerInterface`, as `AbstractExecutor` expects the `purger` to be an instance of `PurgerInterface`, which is not guaranteed with the changes as is: https://github.com/doctrine/data-fixtures/blob/265b209150ce6bfa22c3825bfcd512f5b5c32d3b/lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php#L26.

```suggestion
interface ORMPurgerInterface extends PurgerInterface
```

alcaeus (changes requested):

lib/Doctrine/Common/DataFixtures/Purger/ORMPurgerInterface.php
kissifrot: That's right, thanks, forgot to add it ^^

kissifrot:
@alcaeus However CI cries about Interface suffix, however all other interfaces are sufixed, what ashould I do?

alcaeus:
Since we can't rename the `ORMPurger` class, it's fine to add another exception for the new `ORMPurgerInterface` in phpcs.xml.dist: https://github.com/doctrine/data-fixtures/blob/265b209150ce6bfa22c3825bfcd512f5b5c32d3b/phpcs.xml.dist#L42.

kissifrot:
> 
> 
> Since we can't rename the `ORMPurger` class, it's fine to add another exception for the new `ORMPurgerInterface` in phpcs.xml.dist:
> 
> https://github.com/doctrine/data-fixtures/blob/265b209150ce6bfa22c3825bfcd512f5b5c32d3b/phpcs.xml.dist#L42

Done :)

alcaeus (approved):

alcaeus:
Thanks @kissifrot!

PR #348, opened 2020-05-25
Allow Php 8

greg0ire:


greg0ire:
Should we wait for mongoDB on this one?

alcaeus:
`ignore-platform-reqs` won't get you far, as it will still fail because of the missing MongoDB extension. Let's allow others to test this library against PHP 8 (since its main functionality is not based on MongoDB), then work on getting our own tests stable.

PR #349, opened 2020-06-25
Allows both doctrine/common 3.x & doctrine/persistence 2.x

fruitwasp:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues |

#### Summary

- Allows doctrine/common 3.x & doctrine/persistence 2.x
- Bumps 2.x of doctrine/common to 2.13 [because of release notes](https://github.com/doctrine/common/releases/tag/3.0.0)

Thoughts?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
-        "doctrine/common": "^2.11",
-        "doctrine/persistence": "^1.3.3"
+        "doctrine/common": "^2.13|^3.0",
+        "doctrine/persistence": "^1.3.3|^2.0"
     },
     "conflict": {
         "doctrine/phpcr-odm": "<1.3.0"
--- a/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Doctrine\Common\DataFixtures;
 
-use Doctrine\Common\Util\ClassUtils;
-use Doctrine\Common\Version;
 use function file_exists;
 use function file_get_contents;
 use function file_put_contents;
@@ -31,10 +29,6 @@ class ProxyReferenceRepository extends ReferenceRepository
      */
     protected function getRealClass($className)
     {
-        if (Version::compare('2.2.0') <= 0) {
-            return ClassUtils::getRealClass($className);
-        }
-
         if (substr($className, -5) === 'Proxy') {
             return substr($className, 0, -5);
         }


greg0ire:
> Thoughts?

I think I already expressed them in another of your PRs, but IMO you should start contributing a maximum of things to the stable branch first. This applies to allowing common 3 and persistence 2 I think.

fruitwasp:
what about allowing doctrine/mongodb-odm 2.1.x? Required to allow doctrine/persistence 2.x (see [https://github.com/doctrine/mongodb-odm/pull/2204](https://github.com/doctrine/mongodb-odm/pull/2204)

Ok to drop doctrine/mongodb-odm 1.x too?

greg0ire:
> what about allowing doctrine/mongodb-odm 2.1.x?

If both versions can be handled by a single version of data-fixtures, why not?

garak:
Related PR in mongodb-odm has been merged

composer.json
```diff
@@ -15,29 +15,27 @@
...
     "suggest": {
-        "alcaeus/mongo-php-adapter": "For using MongoDB ODM with PHP 7",
         "doctrine/mongodb-odm": "For loading MongoDB ODM fixtures",
         "doctrine/orm": "For loading ORM fixtures",
         "doctrine/phpcr-odm": "For loading PHPCR ODM fixtures"
     },
     "config": {
         "platform": {
-            "ext-mongo": "1.6.16"
+            "ext-mongodb": "1.6.0"
```
alcaeus, line 38: This can be dropped once you're no longer using mongo-php-adapter.

composer.json
```diff
@@ -15,29 +15,27 @@
     ],
     "require": {
         "php": "^7.2",
-        "doctrine/common": "^2.11",
-        "doctrine/persistence": "^1.3.3"
+        "doctrine/common": "^2.13 | ^3.0",
+        "doctrine/persistence": "^2.0"
```
alcaeus, line 19: This library should not switch to 2.0, but rather `^1.3 | ^2`.

lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
```diff
@@ -140,5 +139,3 @@ public function purge()
      */
     abstract public function execute(array $fixtures, $append = false);
 }
-
-interface_exists(ObjectManager::class);
```
alcaeus, line 144: IIRC the compatibility layer has to stay, care to weigh in @greg0ire?

tests/Doctrine/Tests/Common/DataFixtures/Purger/MongoDBPurgerTest.php
```diff
@@ -31,10 +32,12 @@ private function getDocumentManager()
...
         $config->setHydratorNamespace('Hydrators');
         $config->setMetadataDriverImpl(AnnotationDriver::create(dirname(__DIR__) . '/TestDocument'));
-        AnnotationDriver::registerAnnotationClasses();
 
         $dm = DocumentManager::create(null, $config);
-        if (! $dm->getConnection()->connect()) {
+
+        try {
+            $dm->getClient()->listDatabases();
+        } catch (ConnectionTimeoutException $exception) {
```
alcaeus, line 40: 👍

lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
stof: if you keep support for doctrine/persistence 1.x (which has the class name BC layer), yes

fruitwasp:
tests seem to fail on registering annotations. The previously used `AnnotationDriver::registerAnnotationClasses` does no longer exist? Anyone got an idea on how to proceed?

greg0ire:
The error message says it  `Call to undefined method MongoDB\Collection::getIndexInfo()`, how do does this relate with registering annotations?

fruitwasp:
Targets the 1.4.x stable branch now. Will bump doctrine/mongodb-odm in another pull request

greg0ire (approved):

composer.json
```diff
@@ -15,8 +15,8 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
-        "doctrine/common": "^2.11",
-        "doctrine/persistence": "^1.3.3"
+        "doctrine/common": "^2.13|^3.0",
+        "doctrine/persistence": "^1.3.3|^2.0"
```
SenseException, line 19: It looks like none of the builds is installing versions fitting to the version constraints `^3.0` and `^2.0`. How can we be sure that the newer versions will work too?

greg0ire: How can they be newer? 3 and 2 are the latest versions…

fruitwasp: Both doctrine/common 3 and doctrine/persistence 2 will not install on dev, because of [doctrine/mongodb-odm](https://github.com/doctrine/mongodb-odm) 1.3.x (latest version is 2.1.x) not having support for doctrine/common 3.

A few options:
1. Add support for doctrine/mongodb-odm 2.x & drop support for doctrine/mongodb-odm 1.3.x in this pull request. They don't go together nicely (tried to in a previous version of this pull request)
2. After this pull request is merged upstream, go for the first option but on the dev branch instead.
2. Replace doctrine/common in doctrine/mongodb-odm 1.3.x with the new doctrine packages like done in 2.x](https://github.com/doctrine/mongodb-odm/commit/21501c42aa50578f2cbc494a55ec4cb2995f20c1) [could something like this?](https://github.com/doctrine/data-fixtures/commit/b528aa964b5965d958f0f533df2ed611520dad5c)

greg0ire: Option 3 sounds best, can you manage that?

fruitwasp: Worth a try ;)

Quite the task tho. A little help would be nice. Several classes (1) are no longer available due to the removal of doctrine/common. They will have to be replaced in a way, for which I've found docs https://www.doctrine-project.org/2018/07/12/common-2-9-and-dbal-2-8-and-orm-2-6-2.html

For example, to get proxying to work, we'll have to backport changes like https://github.com/doctrine/mongodb-odm/pull/1875 This results in BC breaks, which I don't think is acceptable on this stable branch.

(1) Classes no longer available;
```
Doctrine\Common\Proxy\AbstractProxyFactory
Doctrine\Common\Proxy\ProxyDefinition
Doctrine\Common\Proxy\ProxyGenerator
Doctrine\Common\Util\ClassUtils
Doctrine\Common\Util\Debug
Doctrine\Common\Proxy\Proxy
```

Will give supporting both doctrine/mongodb-odm 1.x & 2.x another try too at the end of the week.

greg0ire: The classes above are still available in common v3. Why not upgrade to common 3 and replacement packages in v1 of `doctrine/mongodb-odm`?

greg0ire: Also, I think there is another solution: allow `doctrine/mongodb-odm` v2 as a dev dependency of this package… I'm not sure why it isn't the case yet, any idea @alcaeus @malarzm ?

greg0ire: @fruitwasp after taking a deeper look to this, I think option 1 might be best in fact… Apparently `doctrine/mongodb-odm` is just a dev dependency, which means people can already install v2 alongside doctrine fixtures

weaverryan:
Hi!

Giving this a gentle bump :). This library (and DoctrineFixturesBundle) is no longer installable out-of-the-box on Symfony 5 apps:

```
composer create-project symfony/skeleton testing_data_fixtures
cd testing_data_fixtures
# will lock doctrine/common on v3
composer require doctrine/doctrine-bundle doctrine/orm

# this will fail because we're locked at doctrine/common v3
composer require doctrine/data-fixtures
```

It looks like there is a MongoDB issue that we're still checking into? Is that all? 

Thanks!

greg0ire:
> It looks like there is a MongoDB issue that we're still checking into? Is that all?

I think this package needs to be made compatible with `doctrine/mongodb-odm` 2, yeah. When that's done, then the contents of this PR should be enough to fix the issue.

alcaeus:
@greg0ire see https://github.com/doctrine/data-fixtures/pull/351.

composer.json
SenseException: With data-fixtures as a dependency in a project, its require-dev parts won't be affected by a v2 constraint. Suggestion 2 would be possible.

greg0ire:
#351 has been merged, let's trigger a new build

composer.json
greg0ire: @SenseException now that #351 is merged, we have the prefer lowest build that uses 1 and 2, while all other builds use 2 and 3.

greg0ire (approved):

fruitwasp:
Looks like doctrine/common 3 & /persistence have been installed in the builds. Great work!

alcaeus (approved):

greg0ire:
Thanks @fruitwasp !

PR #352, opened 2020-10-01
Compatibility with recent PHPUnit (8)

DavidPrevot:


SenseException:
The current Composer constraint keeps the PHPUnit version to 7.x.

DavidPrevot:
I don’t believe this improvement is incompatible with PHPUnit 7 (the tests run in travis seem to match that belief).

SenseException:
True, it isn't incompatible, but currently not necessary. The minimum supported PHP version of this project is 7.2 though, so you can make a whole PHPUnit update of your PR to version 8 instead of just introducing one return type. You would contribute much more value to the project with it.

While at it, you can also introduce ` : void` to the test methods and add other return types and typed properties to the methods of test classes.

Please also follow the Doctrine coding style: `There must be exactly 1 whitespace between closing parenthesis and return type colon.`. The build of your PR failed because of this.

greg0ire:
Fixed in https://github.com/doctrine/data-fixtures/commit/5e5b0708013d6726ed0e996b116746458b610231

PR #353, opened 2020-12-04
Migrate from Travis CI to GitHub Actions

nicwortel:
Travis CI builds are no longer working, so in order to have CI builds again this repository has to be migrated to GitHub Actions.

Also see https://github.com/orgs/doctrine/projects/7#card-50248167, https://github.com/doctrine/event-manager/pull/34.


nicwortel:
Removing draft status to see if that triggers the build.

greg0ire:
I think at least one Github Action must exist before they can appear, let me set one up real quick.

nicwortel:
Ah, I didn't know that! Thanks 🙂

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,95 @@
...
+          ini-values: "zend.assertions=1"
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
```
nicwortel, line 60: I found this more readable than manually composing Composer commands with and without `--prefer-lowest`, [as is done in doctrine/orm](https://github.com/doctrine/orm/blob/242cf1a33df1b8bc5e1b86c3ebd01db07851c833/.github/workflows/continuous-integration.yml#L49-L55) for instance. What do you think?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: "ramsey/composer-install@v1"
+        with:
+          dependency-versions: "${{ matrix.dependencies }}"
```
nicwortel, line 63: I found this more readable than manually composing Composer commands with and without `--prefer-lowest`, [as is done in doctrine/orm](https://github.com/doctrine/orm/blob/242cf1a33df1b8bc5e1b86c3ebd01db07851c833/.github/workflows/continuous-integration.yml#L49-L55) for instance. What do you think?

greg0ire: I think it looks great, didn't know about that action!
The README says

> It installs your Composer dependencies and caches them for improved build times.

Does that mean the cache step should be removed?

nicwortel:
I decided to introduce PHPStan in a separate PR, as this project hasn't been analysed with PHPStan yet and results in errors even on level 0. Some of these errors appear to be solvable by installing `doctrine/phpcr-odm` as a dev-dependency, but I think the amount of changes required is high enough to justify a separate PR.

.github/workflows/continuous-integration.yml
greg0ire: > There is no need to set up a separate caching step since ramsey/composer-install handles this for you.

It clearly needs to be removed.

greg0ire:
The goal of this PR should be to reproduce what's currently done on Travis so that entirely fine 👍

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,95 @@
...
+          coverage: "xdebug"
+          ini-values: "zend.assertions=1"
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
```
greg0ire, line 59:
```suggestion
```

.github/workflows/continuous-integration.yml
greg0ire: Please open a PR against https://github.com/doctrine/.github/ so that we can discuss this at the org-wide level. Please link add a link to this PR to demonstrate how much it improves the situation.

nicwortel: Nice, I missed that!

If you like, I can use this action in all workflows in this and subsequent PRs.

nicwortel: > Please open a PR against https://github.com/doctrine/.github/ so that we can discuss this at the org-wide level. Please link add a link to this PR to demonstrate how much it improves the situation.

Done: https://github.com/doctrine/.github/pull/16

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: "Install PHP with Xdebug"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          extensions: "mongodb, sqlite3"
+          coverage: "xdebug"
+          ini-values: "zend.assertions=1"
+
+        # Includes caching of dependencies
```
nicwortel, line 53: Do you think this is a useful comment, or should I remove it?

.github/workflows/continuous-integration.yml
greg0ire: > If you like, I can use this action in all workflows in this and subsequent PRs.

I would like that, but let's see if the others would like that too.

.github/workflows/continuous-integration.yml
greg0ire: I think it can safely be removed.

.github/workflows/continuous-integration.yml
nicwortel: Resolving as https://github.com/doctrine/.github/pull/16 has been approved.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP with Xdebug"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          extensions: "mongodb, sqlite3"
+          coverage: "xdebug"
```
greg0ire, line 50:
```suggestion
          coverage: "pcov"
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,88 @@
...
+          - php-version: "7.2"
+            dependencies: "lowest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP with Xdebug"
```
greg0ire, line 45:
```suggestion
      - name: "Install PHP"
```

greg0ire:
The PHPUnit constraint should be updated to `^8.0` so that we can use PCOV

nicwortel:
@greg0ire the build fails after applying your suggestion, I believe PCOV is only supported by PHPUnit 8 and higher.

tests/Doctrine/Tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
```diff
@@ -22,7 +22,7 @@ class ProxyReferenceRepositoryTest extends BaseTest
     public const TEST_ENTITY_ROLE = Role::class;
     public const TEST_ENTITY_LINK = Link::class;
 
-    public static function setUpBeforeClass()
+    public static function setUpBeforeClass(): void
```
greg0ire, line 25:
```suggestion
    public static function setUpBeforeClass() : void
```

greg0ire:
> I'll have to upgrade PHPUnit and add return types to some methods.

That was not a lot of methods, luckily :)

greg0ire:
I think we're good code-wise. Some commits look like they can be fixed up into previous ones (for instance the XDebug to PCOV one), can you please take a look?

nicwortel:
> I think we're good code-wise. Some commits look like they can be fixed up into previous ones (for instance the XDebug to PCOV one), can you please take a look?

I've rebased the branch and merged some commits together.

greg0ire (approved):

greg0ire:
Thanks again @nicwortel ! Great job!

PR #355, opened 2020-12-08
Upgrade the Doctrine Coding Standard

nicwortel:
See https://github.com/doctrine/event-manager/pull/33, https://github.com/orgs/doctrine/projects/7#card-50248429.

In order to support PHP 8, the Coding Standard repository needs to be updated.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -192,7 +192,7 @@ class ORMPurger implements PurgerInterface
                 }
 
                 /** @var ClassMetadata $targetClass */
-                $targetClass = $em->getClassMetadata($assoc['targetEntity']);
+                $targetClass     = $em->getClassMetadata($assoc['targetEntity']);
                 $targetClassName = $targetClass->getName();
 
                 if (! $sorter->hasNode($targetClassName)) {
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -252,14 +252,13 @@ class ORMPurger implements PurgerInterface
     }
 
     /**
-     * @param array            $association
-     * @param ClassMetadata    $class
-     * @param AbstractPlatform $platform
-     *
-     * @return string
+     * @param mixed[] $assoc
      */
-    private function getJoinTableName($assoc, $class, $platform)
-    {
+    private function getJoinTableName(
+        array $assoc,
+        ClassMetadata $class,
+        AbstractPlatform $platform
+    ): string {
         if (isset($assoc['joinTable']['schema']) && ! method_exists($class, 'getSchemaName')) {
             return $assoc['joinTable']['schema'] . '.' . $this->em->getConfiguration()->getQuoteStrategy()->getJoinTableName($assoc, $class, $platform);
         }
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -9,6 +9,7 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
+
 use function array_reverse;
 use function array_search;
 use function count;
@@ -118,7 +119,8 @@ class ORMPurger implements PurgerInterface
         for ($i = count($commitOrder) - 1; $i >= 0; --$i) {
             $class = $commitOrder[$i];
 
-            if ((isset($class->isEmbeddedClass) && $class->isEmbeddedClass) ||
+            if (
+                (isset($class->isEmbeddedClass) && $class->isEmbeddedClass) ||
                 $class->isMappedSuperclass ||
                 ($class->isInheritanceTypeSingleTable() && $class->name !== $class->rootEntityName)
             ) {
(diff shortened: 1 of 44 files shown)


lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -190,7 +192,7 @@ private function getCommitOrder(EntityManagerInterface $em, array $classes)
                 }
 
                 /** @var ClassMetadata $targetClass */
-                $targetClass     = $em->getClassMetadata($assoc['targetEntity']);
+                $targetClass = $em->getClassMetadata($assoc['targetEntity']);
```
greg0ire, line 195: Both of these should have been fixed by `phpcbf` I think :thinking:

greg0ire:
You can add `<config name="php_version" value="70200"/>` towards the top, it might be helpful in avoiding some unwanted fixes.

phpcs.xml.dist
```diff
@@ -14,13 +14,24 @@
...
+        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingTraversableTypeHintSpecification"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingTraversableTypeHintSpecification"/>
 
         <!-- Will cause BC breaks to method signatures - disabled for now -->
-        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint" />
-        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint" />
+        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint"/>
+        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint"/>
+
+        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingAnyTypeHint"/>
```
nicwortel, line 25: There are 27 missing parameter type hints. I think it is out of scope for this PR to fix them all. I suggest creating an issue to track this.

phpcs.xml.dist
```diff
@@ -14,13 +14,24 @@
...
+        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingAnyTypeHint"/>
```
nicwortel, line 26: There are 54 missing return type hints. I think it is out of scope for this PR to fix them all. I suggest creating an issue to track this.

phpcs.xml.dist
```diff
@@ -14,13 +14,24 @@
...
+        <exclude name="SlevomatCodingStandard.PHP.RequireExplicitAssertion.RequiredExplicitAssertion"/>
```
nicwortel, line 27: This sniff wants to automatically replace doc blocks with `assert` calls. This felt a little bit risky because this may actually break things.

nicwortel:
Currently blocked because ODM doesn't allow PHP 8 yet, see https://github.com/doctrine/mongodb-odm/pull/2248.

phpcs.xml.dist
greg0ire: I think I will rather take the time to contribute them to your PR. Feel free to start a separate PR before I do that though. I'd rather not commit "todo" things in the codebase.

greg0ire:
> Currently blocked because doctrine/mongodb-odm doesn't allow PHP 8 yet, see doctrine/mongodb-odm#2248.

Then let's focus on upgrading phpcs and only that, shall we?

nicwortel:
> Then let's focus on upgrading phpcs and only that, shall we?

I have changed the PR title and removed the last commit. The build is running again, but should turn green in a minute.

greg0ire (changes requested):
I'm going to checkout your PR and add the missing type hints

phpcs.xml.dist
greg0ire: Not really, `assert` is not executed at all on sensibly configured production environments.

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -199,6 +199,8 @@ public function getReferenceNames($reference)
      * Checks if reference has identity stored
      *
      * @param string $name
+     *
+     * @return bool
```
nicwortel, line 203: Any reasons for not using the native `: bool` return type?

tests/Doctrine/Tests/Common/DataFixtures/TestEntity/Link.php
```diff
@@ -32,17 +32,17 @@ public function __construct(Uuid $id)
...
     }
 
-    public function getId()
+    public function getId(): Uuid
     {
         return $this->id;
     }
 
-    public function getUrl()
+    public function getUrl(): string
```
nicwortel, line 40: `url` could actually be `null` because it's not initialized in the constructor.

But this is just a test fixture so I'm not sure how important this is.

tests/Doctrine/Tests/Common/DataFixtures/TestTypes/UuidType.php
```diff
@@ -12,25 +12,38 @@ class UuidType extends Type
...
         $fieldDeclaration['length'] = 36;
         $fieldDeclaration['fixed']  = true;
 
         return $platform->getVarcharTypeDeclarationSQL($fieldDeclaration);
     }
 
+    /**
+     * @param ?string $value
+     *
+     * @return ?Uuid
```
nicwortel, line 29: It should be possible to add this as a native return type hint.

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
greg0ire: Yes: the class is not `final`, so that would be a BC-break.

nicwortel: Good reason :smile:

greg0ire (approved):

PR #362, opened 2021-04-21
Remove calls to deprecated methods

bcremer:
Remove call to methods that have been deprecated in Doctrine DBAL.


composer.json
```diff
@@ -16,6 +16,7 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "doctrine/common": "^2.13|^3.0",
+        "doctrine/dbal": "^2.13|^3.0",
```
bcremer: Classes from the `doctrine/dbal` namespace are used directly in the `lib/Doctrine/Common/DataFixtures` so it should be in `require` and not `require-dev`.

lib/Doctrine/Common/DataFixtures/Exception/CircularReferenceException.php
```diff
@@ -4,8 +4,8 @@
 
 namespace Doctrine\Common\DataFixtures\Exception;
 
-use Doctrine\Common\CommonException;
```
bcremer: That would be a BC break. I could revert that if the remaining changes are good for a minor release.

composer.json
greg0ire: Are you sure? I can see `doctrine/mongodb-odm` in the dev dependencies. Should a MongoDB project really install `doctrine/dbal`?

stof: the DBAL classes are used only in the DBAL-specific code, which is optional in the library. So this should indeed be reverted (using a conflict rule to specify that the library is incomaptible with DBAL <2.13 instead)

lib/Doctrine/Common/DataFixtures/Exception/CircularReferenceException.php
greg0ire: Please do revert it :pray: Also, why did you need to do that in the first place, if you recall?

bcremer: I removed the reference to `Doctrine\Common\CommonException` because this class is also deprecated:

> Doctrine\Common\CommonException is deprecated without replacement.
see: https://www.doctrine-project.org/2018/07/12/common-2-9-and-dbal-2-8-and-orm-2-6-2.html

I'm fine to revert that line and just keep the updated DBAL calls.

bcremer:
Which branch should I target for? Still 1.5.x or up to 1.6.x?

greg0ire:
You can stay on 1.5.x IMO, it's not a bugfix technically, but the changes are small enough that it won't break anything (if you do the revert).

lib/Doctrine/Common/DataFixtures/Exception/CircularReferenceException.php
bcremer: Reverted.

bcremer:
Removed the changed exception and rebased to the current `1.5.x`-branch.

composer.json
derrabus: This line still needs to be reverted.

derrabus (changes requested):

composer.json
bcremer: Removed.

greg0ire (approved):

derrabus (approved):

greg0ire:
Thanks @bcremer !

PR #366, opened 2021-09-18
Fix DBAL 3 compatibility

### LINKED ISSUE #365: Loading fixtures fails with DBAL 3
Using DBAL 3, the command `doctrine:fixtures:load` fails with this error:

```
In ORMPurger.php line 133:

  [Symfony\Component\ErrorHandler\Error\UndefinedMethodError]
  Attempted to call an undefined method named "getFilterSchemaAssetsExpression" of class "Doctrine\DBAL\Configuration".
```

https://github.com/doctrine/data-fixtures/blob/51d3d4880d28951fff42a635a2389f8c63baddc5/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php#L133

`Configuration::getFilterSchemaAssetsExpression()` has been [removed](https://github.com/doctrine/dbal/blob/3.1.x/UPGRADE.md#bc-break-removed-previously-deprecated-features) in DBAL 3.

williarin:
Fixes #365 

`doctrine:fixtures:load` command no longer failing with DBAL 3.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -130,7 +130,10 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         }
 
         $connection            = $this->em->getConnection();
-        $filterExpr            = $connection->getConfiguration()->getFilterSchemaAssetsExpression();
+        $filterExpr            = method_exists(
+            $connection->getConfiguration(),
+            'getFilterSchemaAssetsExpression'
+        ) ? $connection->getConfiguration()->getFilterSchemaAssetsExpression() : null;
         $emptyFilterExpression = empty($filterExpr);
 
         $schemaAssetsFilter = method_exists(
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "require-dev": {
         "ext-sqlite3": "*",
         "doctrine/coding-standard": "^9.0",
-        "doctrine/dbal": "^2.5.4",
+        "doctrine/dbal": "^2.5.4 || ^3.0",
         "doctrine/mongodb-odm": "^1.3.0 || ^2.0.0",
         "doctrine/orm": "^2.7.0",
         "phpunit/phpunit": "^8.0"


lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -130,7 +130,7 @@ public function purge()
         }
 
         $connection            = $this->em->getConnection();
-        $filterExpr            = $connection->getConfiguration()->getFilterSchemaAssetsExpression();
+        $filterExpr            = method_exists($connection->getConfiguration(), 'getFilterSchemaAssetsExpression') ? $connection->getConfiguration()->getFilterSchemaAssetsExpression() : '';
```
stof: the fallback should be `null` rather than the empty string, to be consistent with what DBAL 2 returns when not using the filter.

williarin: Oops, indeed.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -130,7 +130,7 @@ public function purge()
         }
 
         $connection            = $this->em->getConnection();
-        $filterExpr            = $connection->getConfiguration()->getFilterSchemaAssetsExpression();
+        $filterExpr            = method_exists($connection->getConfiguration(), 'getFilterSchemaAssetsExpression') ? $connection->getConfiguration()->getFilterSchemaAssetsExpression() : null;
```
greg0ire: Please use linebreaks :pray:

williarin: I think line breaks will not be coherent with the coding style of this file and I'm not sure it improves readability.
It would look like this:
```php
$connection            = $this->em->getConnection();
$filterExpr            = method_exists($connection->getConfiguration(), 'getFilterSchemaAssetsExpression')
    ? $connection->getConfiguration()->getFilterSchemaAssetsExpression()
    : null;
$emptyFilterExpression = empty($filterExpr);

$schemaAssetsFilter = method_exists($connection->getConfiguration(), 'getSchemaAssetsFilter') ? $connection->getConfiguration()->getSchemaAssetsFilter() : null;
```

Maybe formatting can be delegated to a separate PR?

greg0ire: The coding style is really defined at the organization level, not at the file level

Breaking inside `method_exists` would result in a better distribution (in your example, the first line still requires horizontal scrolling).

```php
$connection            = $this->em->getConnection();
$filterExpr            = method_exists(
    $connection->getConfiguration(),
    'getFilterSchemaAssetsExpression'
) ? $connection->getConfiguration()->getFilterSchemaAssetsExpression() : null;
$emptyFilterExpression = empty($filterExpr);

$schemaAssetsFilter = method_exists(
    $connection->getConfiguration(),
    'getSchemaAssetsFilter'
) ? $connection->getConfiguration()->getSchemaAssetsFilter() : null;
```

> Maybe formatting can be delegated to a separate PR?

Sure! See https://github.com/doctrine/data-fixtures/pull/367

greg0ire:
To make this complete, I think we should allow dbal 3 the version constraint here: https://github.com/doctrine/data-fixtures/blob/d5ebdb48cdfb2ac2fad612f339f568c098b87caf/composer.json#L27

Currently, we are only testing with `doctrine/dbal` v2 because of it.

greg0ire (dismissed):

greg0ire:
Maybe I approved too fast… what does this PR do?
I think it means that when upgrading to DBAL 3, tables that were previously not purged will suddenly be purged, correct?

EDIT: and then I click "Close and comment" when I mean to click comment :sweat_smile:

stof:
> Maybe I approved too fast… what does this PR do?
> I think it means that when upgrading to DBAL 3, tables that were previously not purged will suddenly be purged, correct?

Well, doctrine/data-fixtures does not *configure* the filtering. It *uses* the filtering configured in DBAL.
If you use the filtering feature of your project, you will already need to migrate to the non-deprecated feature before migrating to DBAL 3.

greg0ire:
Oh right, I didn't notice that the new way was already handled in the lines below (because I didn't know what it looked like). It should be fine then I think.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @williarin !

PR #380, opened 2022-03-21
Document possible incompatibility

### LINKED ISSUE #364: There is no active transaction
PHP version: 8.0.6
MySQL version: 5.7

After updating php from 7.0 to 8.0 this issue arises. All the fixtures successfully load their data into database, but after that process fails.

image

I think this is related to PHP's BC break:
>  PDO::inTransaction() now reports the actual transaction state of the connection, rather than an approximation maintained by PDO. If a query that is subject to "implicit commit" is executed, PDO::inTransaction() will subsequently return false, as a transaction is no longer active.

greg0ire:
People are getting confused when getting "There is no active
transaction" from MySQL.

Closes #364

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,25 @@ class MyOtherFixture extends AbstractFixture
 
 **Notice** the ordering is relevant to Loader class.
 
+## Atomicity
+
+This package exposes an `AbstractExecutor` class and 3 concrete implementations
+for `doctrine/orm`, `doctrine/mongodb-odm` and `doctrine/phpcr-odm`.
+
+The executors purge the database, then load the fixtures. The ORM
+implementation wraps these two steps in a database transaction, that way the
+loading either succeeds or fails cleanly, meaning nothing is actually changed
+in the database at the end of the loading. It delegates the purging to a
+separate class that can be configured to either use a `TRUNCATE` or a `DELETE`
+statement to empty tables.
+
+Not all RDBMS have the capability to allow `TRUNCATE` statements inside
+transactions though. Notably, MySQL will produce the infamous "There is no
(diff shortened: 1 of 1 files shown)


greg0ire:
@rtek please review.

README.md
```diff
@@ -190,6 +190,25 @@ class MyOtherFixture extends AbstractFixture
...
 
+## Atomicity
+
+This package exposes an `AbstractExecutor` class and 3 concrete implementations
+for `doctrine/orm`, `doctrine/mongodb-odm` and `doctrine/phpcr-odm`.
+
+The executors purge the database, then load the fixtures. The ORM
+implementation wraps these two steps in a database transaction, that way the
+loading either succeeds or fails cleanly, meaning nothing is actually changed
+in the database at the end of the loading. It delegates the purging to a
```
rtek, line 201: >that way the loading either succeeds or fails cleanly, meaning nothing is actually changed in the database ~at the end of the loading~ if the load fails.

rtek (changes requested):

README.md
greg0ire: Do you mean "if the loading fails"?

rkorczykowski: yessir other than that its good

rtek: yessir other than that its good.

derrabus (approved):

SenseException (approved):

README.md
greg0ire: Woops forgot to push the fix :man_facepalming: … that will be for another PR

greg0ire: https://github.com/doctrine/data-fixtures/pull/381

PR #384, opened 2022-03-26
Avoid inline literals in titles

greg0ire:
They result in weird links weird inline literals are replaced with
hashes.
Let us also add missing words in sentences above, this should make it
easier to figure out that these are the same 2 options as below, or just
make sentences correct.

Screenshot 2022-03-26 at 12-21-17 Doctrine Data Fixtures - Doctrine Data fixtures

I had to fiddle with CSS properties to make these links appear. Maybe this fix will result in the following weird behavior to disappear as well?

Peek 2022-03-26 11-51

### DIFF
--- a/docs/en/how-to/loading-fixtures.rst
+++ b/docs/en/how-to/loading-fixtures.rst
@@ -14,8 +14,8 @@ Creating a fixture class
 Fixture classes have two requirements:
 
 * They must implement ``Doctrine\Common\DataFixtures\FixtureInterface``.
-* If they have a constructor, that constructor should not accept any
-  arguments.
+* If they have a constructor, that constructor should be invokable
+  without arguments.
 
 .. code-block:: php
     <?php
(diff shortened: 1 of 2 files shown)


greg0ire:
Note that the first bug looks like it should be fixed by https://github.com/doctrine/rst-parser/pull/135 or https://github.com/doctrine/rst-parser/pull/124, but doctrine-website seems to be using a very outdated version of `doctrine/rst-parser`

SenseException:
AFAIK the outdated version is used because of the buggy table rendering in the past that broke the builds. I started with an update for the website for using at least PHP 8, but some dependencies need to be updated too. I'm currently not able to work much in a week on different construction zones.

docs/en/how-to/loading-fixtures.rst
```diff
@@ -12,8 +12,9 @@ Creating a fixture class
 ------------------------
 
 Fixture classes have two requirements:
-- They must implement ``Doctrine\Common\DataFixtures\FixtureInterface``.
-- If they have a constructor, that constructor should not accept any
+
+* They must implement ``Doctrine\Common\DataFixtures\FixtureInterface``.
+* If they have a constructor, that constructor should not accept any
   arguments.
```
derrabus:
```suggestion
* If they have a constructor, that constructor should be invokable
  without arguments.
```

PR #388, opened 2022-04-20
Add singleTransaction option for ORMExecutor

BafS:
Reworked https://github.com/doctrine/data-fixtures/pull/209 to target 1.6.x ~~and use `func_get_args()` to avoid BC~~ use the constructor to avoid BC


lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
```diff
@@ -144,10 +144,11 @@ public function purge()
...
      * Executes the given array of data fixtures.
      *
-     * @param array $fixtures Array of fixtures to execute.
-     * @param bool  $append   Whether to append the data fixtures or purge the database before loading.
+     * @param FixtureInterface[] $fixtures          Array of fixtures to execute.
+     * @param bool               $append            Whether to append the data fixtures or purge the database before loading.
+     * @param bool               $singleTransaction Whether to use a single transaction when loading fixtures.
      */
-    abstract public function execute(array $fixtures, $append = false);
+    abstract public function execute(array $fixtures, $append = false /* bool $singleTransaction = false */);
```
stof: instead of adding a new argument that is relevant only for the ORM, maybe this should be a constructor argument of the ORMExecutor instead.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -62,17 +62,33 @@ public function setReferenceRepository(ReferenceRepository $referenceRepository)
...
+                }
+
+                foreach ($fixtures as $fixture) {
+                    $executor->load($em, $fixture);
+                }
+            });
+        } else {
             if ($append === false) {
-                $executor->purge();
+                $this->em->beginTransaction();
```
stof: use `$this->em->getConnection()->transactional()` instead. this will ensure handling properly the rollback on exceptions.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -62,17 +62,33 @@ public function setReferenceRepository(ReferenceRepository $referenceRepository)
...
+                $this->purge();
+                $this->em->commit();
             }
 
             foreach ($fixtures as $fixture) {
-                $executor->load($em, $fixture);
+                $this->em->beginTransaction();
```
stof: The usage of `EntityManager::transactional()` was adding an implicit `flush` that you removed. That's a BC break.

BafS: I didn't change this part but I'll use the closure style everywhere

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -20,19 +20,23 @@ class ORMExecutor extends AbstractExecutor
     /** @var ORMReferenceListener */
     private $listener;
 
+    /** @var bool */
+    private $singleTransaction;
```
BafS: Can I use `private bool $singleTransaction;`?

greg0ire: Nope: this package still supports PHP 7.2, and typed properties are a 7.4 feature.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
stof: well, you *did* change that part, by switching from `EntityManager::transactional()` to using `beginTransaction`

stof:
This PR needs tests covering the case of not using a single transaction

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
BafS: Right!

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
BafS: `$singleTransaction` would be true by default and I kept the same behavior there to avoid BC, I used `beginTransaction` for the new part only (the else), maybe I'm missing your point but I just indented the previous code to have it under the if, the git diff doesn't help.

BafS:
> This PR needs tests covering the case of not using a single transaction

Thanks, I added some tests, let me know if it's enough.

stof:
The test is exactly the same than the test not using multiple transactions. So to me, this means that the test has no value. If I revert all changes in the `ORMExecutor::execute` method, your test would still be passing.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
BafS: So now the code is using `$this->em->wrapInTransaction()`...

BafS:
I rebased my PR to use `$this->em->wrapInTransaction()` instead of `$this->em->getConnection()->transactional()` and added some tests to count the number of "transactional" calls

greg0ire:
Please add documentation about the new feature.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -77,14 +81,29 @@ public function setReferenceRepository(ReferenceRepository $referenceRepository)
...
+                    $executor->purge();
+                });
             }
 
+            $em = $this->em;
             foreach ($fixtures as $fixture) {
-                $executor->load($em, $fixture);
+                $this->em->wrapInTransaction(static function () use ($executor, $fixture, $em) {
+                    $executor->load($em, $fixture);
+                });
```
derrabus:
```suggestion
                $this->em->wrapInTransaction(static function (EntityManagerInterface $em) use ($executor, $fixture) {
                    $executor->load($em, $fixture);
                });
```

`wrapInTransaction()` will pass the EM to your closure.

BafS: Good to know, thank you

docs/en/how-to/loading-fixtures.rst
```diff
@@ -106,3 +106,11 @@ then pass ``true`` to the 2nd argument of execute:
 .. code-block:: php
     <?php
     $executor->execute($loader->getFixtures(), append: true);
+
+By default the ``ORMExecutor`` will wrap the purge and the load of fixtures
+in a single transaction. You can pass ``false`` to the third argument
+to wrap the purge and the load of every fixture in its own transaction:
```
greg0ire: As a user, why would I want to do that?

BafS: I updated the text with an example, I don't know if you had another example in mind but that was the issue for me.

docs/en/how-to/loading-fixtures.rst
```diff
@@ -106,3 +106,13 @@ then pass ``true`` to the 2nd argument of execute:
...
+
+By default the ``ORMExecutor`` will wrap the purge and the load of fixtures
+in a single transaction, which is the recommended way, but in some cases (for
+example if loading your fixtures is too slow and causes timeouts) you may
+want to wrap the purge and the load of every fixture in its own transaction.
+To do so, you can pass ``false`` to the third argument:
+
+.. code-block:: php
+    <?php
+    $executor->execute($loader->getFixtures(), singleTransaction: false);
```
derrabus: This does not look right. `$singleTransaction` is a constructor parameter of `ORMExecutor`, not a parameter of the `execute()` method.

derrabus (changes requested):

docs/en/how-to/loading-fixtures.rst
BafS: My bad! Thanks for the notice

docs/en/how-to/loading-fixtures.rst
greg0ire: > I don't know if you had another example in mind

I didn't have any reason in mind, I don't need that change myself. To be clear, your issue would be solved by disabling transactions entirely, but you'd rather have transactions on a fixture per fixture basis because… it's easier to start from there I suppose? The only advantage I can see is that will have fewer things to cleanup (you won't have to cleanup anything for the fixture that failed).

docs/en/how-to/loading-fixtures.rst
```diff
@@ -106,3 +106,13 @@ then pass ``true`` to the 2nd argument of execute:
 .. code-block:: php
     <?php
     $executor->execute($loader->getFixtures(), append: true);
+
+By default the ``ORMExecutor`` will wrap the purge and the load of fixtures
+in a single transaction, which is the recommended way, but in some cases (for
+example if loading your fixtures is too slow and causes timeouts) you may
+want to wrap the purge and the load of every fixture in its own transaction.
+To do so, you can pass ``false`` to the third argument of ``ORMExecutor``:
```
derrabus: This is a bit odd. You talk about a third argument, but the code uses named arguments. Maybe we should rephrase that.

```suggestion
To do so, construct ``ORMExecutor`` with ``singleTransaction: false``:
```

derrabus (approved):
LGTM, just some nit-picking.

docs/en/how-to/loading-fixtures.rst
BafS: I think it's good to still have transaction per fixture in case of failure, as you said, and yes, it's mostly because it's easier to have a flag to change this instead of disabling the transactions entirely (typically in a CI context).

docs/en/how-to/loading-fixtures.rst
BafS: I agree, I did it like this because of the previous sentence, shall I change it too?

> If you want to append the fixtures instead of purging before loading
> then pass ``true`` to the 2nd argument of execute:

docs/en/how-to/loading-fixtures.rst
greg0ire: So if there was a flag to disable transactions entirely, that would fix your issue as well, right? I wonder if using a boolean here is the right thing to do. I'm foreseeing issues where people have their issues fixed by disabling transactions entirely, but not by using a `per-fixture` mode.  Maybe we should have this instead:

```php
/* Convert this to enum when dropping support for PHP 7 */
interface TransactionMode
{
    public const SINGLE = 'single';
    public const PER_FIXTURE = 'per_fixture';
    public const DISABLED = 'disabled';
}
```

I'm not saying you are supposed to implement all modes, but I think using an enum-like will leave some room for additional modes if needed. Also, I think you would fix more issues by implementing `DISABLED` rather than `PER_FIXTURE`.

BafS: That would also solve the issue yes, what I like about having one transaction per fixture is that it's often pretty close to the business logic in HTTP context, while a single transaction for everything will never happen.

docs/en/how-to/loading-fixtures.rst
derrabus: If you'd like to, yes. 🙂

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -25,12 +25,14 @@ class ORMExecutor extends AbstractExecutor
...
+
     /**
-     * Construct new fixtures loader instance.
-     *
-     * @param EntityManagerInterface $em EntityManagerInterface instance used for persistence.
+     * @param EntityManagerInterface $em                EntityManagerInterface instance used for persistence.
+     * @param bool                   $singleTransaction Whether to use a single transaction when loading fixtures.
      */
-    public function __construct(EntityManagerInterface $em, ?ORMPurgerInterface $purger = null)
+    public function __construct(EntityManagerInterface $em, ?ORMPurgerInterface $purger = null, bool $singleTransaction = true)
```
greg0ire: Let's avoid a boolean here please. We've seen that there could be at least one more strategy (no transaction at all), so this is not future-proof. Let's maybe instead inject an interface, and write 2 implementations of that interface? If that's too complicated, let's go with a string that should be part of an enum-like constant.

greg0ire (changes requested):

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
derrabus: > Let's maybe instead inject an interface, and write 2 implementations of that interface?

Before we do that, let's rather do a second `ORMExecutor`. That interface you're suggesting would basically contain the main method of `AbstractExecutor`, `execute()`.

greg0ire: Even better 👍

BafS: Something like `MultipleTransactionsORMExecutor`? There would be quite a lot in common with the `ORMExecutor` and AFAIK we don't want to use inheritance.
Why not having an `ExecutorInterface` as a strategy in the ORMExecutor?

greg0ire: Hard to figure out what you mean exactly, can you push something to this PR?

BafS: I pushed a PoC, it's a bit in the idea of having an enum-like while keeping the logic in classes.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -45,6 +48,8 @@ public function __construct(EntityManagerInterface $em, ?ORMPurgerInterface $pur
             $this->purger->setEntityManager($em);
         }
 
+        $this->executorStrategy = $executorStrategy ? new $executorStrategy($this) : new SingleTransactionStrategy($this);
```
greg0ire: This is a circular dependency. A better design might be the following:

```php
<?php
interface Executor
{
}

abstract class AbstractExecutor implements Executor
{
}

class ORMExecutor extends AbstractExecutor
{
}

class MultipleTransactionORMExecutor implements Executor
{
   public function __construct(private readonly ORMExecutor $executor)
   {}


   {
        $executor = $this->executor;
        if ($append === false) {
            $em->wrapInTransaction(static function () use ($executor) {
                $executor->purge();
            });
        }

            $em->wrapInTransaction(static function (EntityManagerInterface $em) use ($executor, $fixture) {
            });
        }
   }
}
```

What do you think?

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
derrabus: This PoC is insanely complex for such a simple change. Please, let's make this a separate executor. Use a trait if you need to share logic between both executors.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
BafS: It looks good too, I updated the PR

greg0ire (approved):
Great job!

lib/Doctrine/Common/DataFixtures/Executor/Executor.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Common\DataFixtures\Executor;
+
+use Doctrine\Common\DataFixtures\FixtureInterface;
+
+interface Executor
```
derrabus: Does this interface still have a purpose?

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
```diff
@@ -60,6 +58,11 @@ public function getObjectManager()
         return $this->originalManager;
     }
 
+    public function getEntityManager(): EntityManagerInterface
+    {
+        return $this->em;
+    }
```
derrabus: I honestly dislike that the `ORMExecutor` is leaking the EM now.

docs/en/how-to/loading-fixtures.rst
```diff
@@ -101,8 +101,18 @@ do the following:
...
+
+By default the ``ORMExecutor`` will wrap the purge and the load of fixtures
+in a single transaction, which is the recommended way, but in some cases (for
+example if loading your fixtures is too slow and causes timeouts) you may
+want to wrap the purge and the load of every fixture in its own transaction.
+To do so, you can use ``MultipleTransactionORMExecutor``.
+
+.. code-block:: php
+    <?php
+    $executor = new MultipleTransactionORMExecutor(new ORMExecutor($entityManager));
```
derrabus: From a user perspective, it feels unnecessarily complicated that I have to nest both executors into each other.

derrabus (changes requested):
Sorry for taking another round, but I'm really not happy with those nested executors.

lib/Doctrine/Common/DataFixtures/Executor/Executor.php
greg0ire: It's new, and I thought it could be useful in case there is a need for external packages to type hint, against something, but maybe introduce it is premature? We might want to rethink the `$append` boolean part.

lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php
greg0ire: Didn't spot that, indeed it's not great… one way to avoid it would be to introduce a `wrapInTransaction` proxy method on `ORMExecutor`

docs/en/how-to/loading-fixtures.rst
greg0ire: There could be a static method that could act as a named constructor to avoid this.

lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php
```diff
@@ -141,12 +141,7 @@ public function purge()
...
     }
 
-    /**
-     * Executes the given array of data fixtures.
-     *
-     * @param array $fixtures Array of fixtures to execute.
-     * @param bool  $append   Whether to append the data fixtures or purge the database before loading.
-     */
+    /** @inheritDoc */
     abstract public function execute(array $fixtures, $append = false);
```
derrabus, line 150: *If* we keep the `Executor` interface, you can remove the `execute()` method here.

greg0ire:
@derrabus If you prefer to have `MultipleTransactionORMExecutor extends AbstracExecutor`, that's fine by me. I think there is a need for a design change here, but it might best be done outside this PR, and go further than what has been done here. Let us know what you think.

lib/Doctrine/Common/DataFixtures/Executor/Executor.php
derrabus: I see, we need it because the new executor does not extend `AbstractExecutor` anymore. 😕

BafS:
I updated the PR to make `MultipleTransactionORMExecutor` extends `ORMExecutor`, which is probably the easiest way to go but not the nicest (because inheritance). If you prefer I can extends the `AbstracExecutor` and share the logic with trait (which is not so nice either IMO). Not trivial to come up with a nice design without BC.

derrabus:
Let me explore that path with the shared trait. I'll get back to you. And sorry for the back-and-forth on this PR. 😓

BafS:
@derrabus Okay 👍  no problem, it's great to have feedback and to see things moving

derrabus:
Here's my take: #396

derrabus:
Closing in favor of #396.

PR #390, opened 2022-04-21
Upgrade PHPUnit to v9

BafS:
Upgrade PHPUnit to v9 and fix `PHPUnit\Framework\TestCase::prophesize()` deprecation


BafS:
My fork was not up to date

PR #393, opened 2022-04-22
Migrate to ORMSetup

derrabus:
This PR switches the test setup to the new `ORMSetup` class introduced with ORM 2.12.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "conflict": {
         "doctrine/dbal": "<2.13",
-        "doctrine/orm": "<2.10",
+        "doctrine/orm": "<2.12",
         "doctrine/phpcr-odm": "<1.3.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


stof:
if you only run tests on doctrine/orm 2.12+, you should add a conflict rule to make sure that this package won't be installed alongside an older version of the orm anymore (as done for older versions of the ODM), to avoid cases where the compat with 2.11 is broken but undetected by the CI.

derrabus:
Fair enough.

stof:
btw, that's the same kind of conflict (but more restrictive) than the one you add in #392 :smile:

derrabus:
I know, but in #392 I've changed production code while this PR changes the test suite only. But your argument about the CI makes total sense.

stof:
The alternative if you don't want to bump the min supported version of the ORM is to make the testsuite switch between both classes based on `class_exists(ORMSetup::class)`

greg0ire (approved):

PR #397, opened 2022-06-16
Address commonexception deprecation

MarcBrillault:


greg0ire (approved):

SenseException (approved):
Looks good, but @greg0ire how are BC breaks handled in this repo that 1.5 to 1.6 introduce them?

greg0ire:
@SenseException Marc is my colleague, and we wrote this PR together. In particular, this sentence: "We don't think anyone catches this exception in a `catch (CommonException)` statement". It's my personal opinion that a BC-break is OK in this case. If you feel otherwise, we can create 2.0.x and retarget to that.

SenseException:
I see data-fixtures mostly in a dev env in projects, which is why I approved the PR. I wanted to be sure that nothing was missed with these changes and that the chosen version was on purpose.

greg0ire:
Thanks @MarcBrillault !

PR #400, opened 2022-07-09
Fixtures loader improvements

misantron:


lib/Doctrine/Common/DataFixtures/Loader.php
```diff
@@ -79,14 +77,14 @@ class Loader
      *
      * @return array $fixtures Array of loaded fixture object instances.
      */
-    public function loadFromDirectory($dir)
+    public function loadFromDirectory(string $dir): array
```
greg0ire: This is a BC break for extending classes and as such, should be contributed to 2.x
I can create that branch if you feel up to that.

lib/Doctrine/Common/DataFixtures/Loader.php
```diff
@@ -206,32 +204,28 @@ public function getFixtures()
      * Check if a given fixture is transient and should not be considered a data fixtures
      * class.
      *
-     * @psalm-param class-string<object> $className
```
greg0ire: Why?

lib/Doctrine/Common/DataFixtures/Loader.php
```diff
@@ -374,20 +370,18 @@ private function orderFixturesByDependencies()
     /**
      * @psalm-param iterable<class-string> $dependenciesClasses
      */
-    private function validateDependencies(iterable $dependenciesClasses): bool
+    private function validateDependencies(array $dependenciesClasses): void
```
greg0ire: Why?

greg0ire:
Please add a description, and use separate commits with good messages for each thing you do.
You can learn more about this in [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).

lib/Doctrine/Common/DataFixtures/Loader.php
misantron: I can rollback methods type hinting to not create breaking changes

lib/Doctrine/Common/DataFixtures/Loader.php
misantron: Because method argument is an array not an iterable and there is no need to return boolean result - just throw an exception when validation is failed.

greg0ire: > Because method argument is an array not an iterable 

What are you talking about? An array is an iterable, so it is both things at once.

lib/Doctrine/Common/DataFixtures/Loader.php
greg0ire: If you want this merged on 2.x, you definitely will have to.

lib/Doctrine/Common/DataFixtures/Loader.php
misantron: I understand that array is iterable, but when `validateDependencies` method is actually calling it always accept an array because `$fixture->getDependencies()` always return an array. It just what I want to said.

greg0ire: Ok, it's always an array… so what? `validateDependencies` does not need it to be an array, it's only using it in a foreach, all it needs it to be is iterable.

misantron: Ok, I just left only changes I wanted to do, without BC breaks

greg0ire:
When I say "use separate commits with good messages for each thing you do" I mean this:

- add return type hints
- remove extension in filename
- configure recursive directory iterator

I don't mean this:

- improve stuff
- revert type hints
- fix cs
- …
- more fix cs

We don't care that you need to fix cs, we want explanations for very different issues you happen to fix in the same files.

PR #404, opened 2022-08-31
Avoid purging unrelated tables

mpdude:
Assume the following class hierarchy:

```php
/** @Entity */
class Parent {
    // @Id and other attributes here
}

/** @Entity */
class Child extends Parent {
    // @Id and other attributes here
}
```

Both entities have their own, completely unrelated database tables.

I think it is ok (at least, I have a project where it seems to work) to have two different Entity Managers configured, and the first one knows only about the `Parent`, the other about the `Child` entity.

When using the `ORMPurger` with the EntityManager for class `Child`, it should not try to truncate/delete the `parent` table. This is what this PR tries to do – do not include another table when we're not dependent on it, when there is no inheritance mapping relationship.

(OTOH, I'm a bit unsure if the approach is sane... Can the `Child` Entity Manager _not_ know about `Parent`? If it needs to know that class, the way the ORMPurger works would be to include that table as well, since the ORMPurge purges all tables managed by a given EntityManager?)

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -179,16 +179,18 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
                 $sorter->addNode($class->name, $class);
             }
 
-            // $class before its parents
-            foreach ($class->parentClasses as $parentClass) {
-                $parentClass     = $em->getClassMetadata($parentClass);
-                $parentClassName = $parentClass->getName();
+            if ($class->isInheritanceTypeJoined() || $class->isInheritanceTypeSingleTable()) {
+                // $class before its parents
+                foreach ($class->parentClasses as $parentClass) {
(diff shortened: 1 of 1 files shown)


stof:
> (OTOH, I'm a bit unsure if the approach is sane... Can the `Child` Entity Manager _not_ know about `Parent`? If it needs to know that class, the way the ORMPurger works would be to include that table as well, since the ORMPurge purges all tables managed by a given EntityManager?)

If an entity extends another one, the EntityManager cannot known only the child class. That's not a supported usage of doctrine/orm (and would probably require weird hacks to even attempt doing that)

mpdude:
Yeah, I became aware of that as I wrote the above description and wondered what I was doing…

mpdude:
Given that the ORMPurger is designed to purge _all_ entity classes in a given EntityManager, and that the EM cannot manage an entity class that inherits from another one without also knowing that parent class (as stof explained above), the current behavior makes sense.

Thank you for the clarification.

PR #405, opened 2022-09-07
Don't purge read-only marked table

Superkooka:
It makes no sense to purge tables marked as read-only


greg0ire:
Thanks @Superkooka !

stof:
This is wrong to me. Doctrine ORM's `readOnly` in metadata does not actually mean that the ORM only reads in it (yes, this naming is crappy). The ORM will happily insert or delete from that table if you use `$em->persist()` or `$em->remove()`. The only thing it will not do is *updating* objects that it has read from the DB.

In a sense, this is similar to readonly properties in PHP: you can still write to them once or unset them.

greg0ire:
Let me revert this then.

PR #407, opened 2022-10-02
Disable foreign key checks on purge for all mysql versions

croensch:
For your consideration, see #272. I don't know if the discussion was a yes/no to this feature.

I can push a little refactor of `getTruncateTableSQL()` to make it fit better.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -288,6 +288,9 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         $sql = $platform->getTruncateTableSQL($tbl, true);
 
         if ($connection->getDriver() instanceof AbstractMySQLDriver) {
+            // MySQL TRUNCATE TABLE Statement: fails for an InnoDB or NDB table
+            // if there are any FOREIGN KEY constraints from other tables that
+            // reference the table.
             $sql = 'SET FOREIGN_KEY_CHECKS = 0;' . $sql . ';SET FOREIGN_KEY_CHECKS = 1;';
         }
 
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -12,6 +12,7 @@ use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
+use PHPUnit\Framework\MockObject\MockObject;
 
 use function array_reverse;
 use function array_search;
@@ -289,7 +290,7 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         return 'DELETE FROM ' . $tableIdentifier->getQuotedName($platform);
     }
 
-    private function disableForeignKeyChecksForMySQL(Connection $connection)
+    private function disableForeignKeyChecksForMySQL(Connection $connection): void
     {
         if ($this->purgeMode !== self::PURGE_MODE_TRUNCATE || ! $this->isMySQL($connection)) {
             return;
@@ -302,7 +303,7 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         $connection->executeStatement('SET FOREIGN_KEY_CHECKS = 0');
     }
 
-    private function enableForeignKeyChecksForMySQL(Connection $connection)
+    private function enableForeignKeyChecksForMySQL(Connection $connection): void
     {
         if ($this->purgeMode !== self::PURGE_MODE_TRUNCATE || ! $this->isMySQL($connection)) {
             return;
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -288,7 +288,7 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         $sql = $platform->getTruncateTableSQL($tbl, true);
 
         if ($connection->getDriver() instanceof AbstractMySQLDriver) {
-            $sql = 'SET FOREIGN_KEY_CHECKS = 0;'.$sql.';SET FOREIGN_KEY_CHECKS = 1;';
+            $sql = 'SET FOREIGN_KEY_CHECKS = 0;' . $sql . ';SET FOREIGN_KEY_CHECKS = 1;';
         }
 
         return $sql;
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -283,14 +283,7 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         return 'DELETE FROM ' . $tableIdentifier->getQuotedName($platform);
     }
 
-    /**
-     * @param AbstractPlatform $platform
-     * @param Connection       $connection
-     * @param string           $tbl
-     *
-     * @return string
-     */
-    private function getTruncateTableSQL(AbstractPlatform $platform, Connection $connection, $tbl)
+    private function getTruncateTableSQL(AbstractPlatform $platform, Connection $connection, string $tbl): string
     {
         $sql = $platform->getTruncateTableSQL($tbl, true);
 
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -6,8 +6,9 @@ namespace Doctrine\Common\DataFixtures\Purger;
 
 use Doctrine\Common\DataFixtures\Sorter\TopologicalSorter;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\AbstractMySQLDriver;
+use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -15,6 +16,7 @@ use Doctrine\ORM\Mapping\ClassMetadata;
 use function array_reverse;
 use function array_search;
 use function assert;
+use function class_exists;
 use function count;
 use function is_callable;
 use function method_exists;
@@ -143,6 +145,8 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
             'getSchemaAssetsFilter'
         ) ? $connection->getConfiguration()->getSchemaAssetsFilter() : null;
 
+        $this->disableForeignKeyChecksForMySQL($connection);
+
         foreach ($orderedTables as $tbl) {
             // If we have a filter expression, check it and skip if necessary
             if (! $emptyFilterExpression && ! preg_match($filterExpr, $tbl)) {
@@ -162,9 +166,11 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
             if ($this->purgeMode === self::PURGE_MODE_DELETE) {
                 $connection->executeStatement($this->getDeleteFromTableSQL($tbl, $platform));
             } else {
-                $connection->executeStatement($this->getTruncateTableSQL($platform, $connection, $tbl));
+                $connection->executeStatement($platform->getTruncateTableSQL($tbl));
             }
         }
+
+        $this->enableForeignKeyChecksForMySQL($connection);
     }
 
     /**
@@ -283,17 +289,37 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         return 'DELETE FROM ' . $tableIdentifier->getQuotedName($platform);
     }
 
-    private function getTruncateTableSQL(AbstractPlatform $platform, Connection $connection, string $tbl): string
+    private function disableForeignKeyChecksForMySQL(Connection $connection)
     {
-        $sql = $platform->getTruncateTableSQL($tbl, true);
-
-        if ($connection->getDriver() instanceof AbstractMySQLDriver) {
-            // MySQL TRUNCATE TABLE Statement: fails for an InnoDB or NDB table
-            // if there are any FOREIGN KEY constraints from other tables that
-            // reference the table.
-            $sql = 'SET FOREIGN_KEY_CHECKS = 0;' . $sql . ';SET FOREIGN_KEY_CHECKS = 1;';
+        if ($this->purgeMode !== self::PURGE_MODE_TRUNCATE || ! $this->isMySQL($connection)) {
+            return;
         }
 
-        return $sql;
+        // see MySQL TRUNCATE TABLE Statement: fails for an InnoDB or NDB table
+        // if there are any FOREIGN KEY constraints from other tables that
+        // reference the table.
+        $connection->executeStatement('SET @DOCTRINE_OLD_FOREIGN_KEY_CHECKS = @@FOREIGN_KEY_CHECKS');
+        $connection->executeStatement('SET FOREIGN_KEY_CHECKS = 0');
+    }
+
+    private function enableForeignKeyChecksForMySQL(Connection $connection)
+    {
+        if ($this->purgeMode !== self::PURGE_MODE_TRUNCATE || ! $this->isMySQL($connection)) {
+            return;
+        }
+
+        $connection->executeStatement('SET FOREIGN_KEY_CHECKS = @DOCTRINE_OLD_FOREIGN_KEY_CHECKS');
+    }
+
+    private function isMySQL(Connection $connection): bool
+    {
+        $platform = $connection->getDatabasePlatform();
+
+        if (! class_exists(AbstractMySQLPlatform::class)) {
+            // before DBAL 3.3
+            return $platform instanceof MySQLPlatform;
+        }
+
+        return $platform instanceof AbstractMySQLPlatform;
     }
 }
--- a/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
+++ b/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
@@ -12,7 +12,6 @@ use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
-use PHPUnit\Framework\MockObject\MockObject;
 
 use function array_reverse;
 use function array_search;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -280,4 +282,22 @@ private function getDeleteFromTableSQL(string $tableName, AbstractPlatform $plat
...
     }
+
+    /**
+     * @param AbstractPlatform $platform
+     * @param Connection       $connection
+     * @param string           $tbl
+     *
+     * @return string
+     */
+    private function getTruncateTableSQL(AbstractPlatform $platform, Connection $connection, $tbl)
```
greg0ire:
```suggestion
    private function getTruncateTableSQL(AbstractPlatform $platform, Connection $connection, string $tbl): string
```

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -280,4 +282,22 @@ private function getDeleteFromTableSQL(string $tableName, AbstractPlatform $plat
...
+    {
+        $sql = $platform->getTruncateTableSQL($tbl, true);
+
+        if ($connection->getDriver() instanceof AbstractMySQLDriver) {
```
greg0ire: A comment explaining in detail why this is needed would be nice.

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,81 @@
...
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Driver\AbstractMySQLDriver;
+use Doctrine\DBAL\Driver\AbstractSQLiteDriver;
+use ReflectionClass;
+
+/**
+ * @author Robert Freigang <robertfreigang@gmx.de>
+ *
+ * @covers ORMPurger::getTruncateTableSQL
```
greg0ire: From https://phpunit.readthedocs.io/en/9.5/annotations.html?highlight=covers#covers:

> Please note that this annotation requires a fully-qualified class name (FQCN). To make this more obvious to the reader, it is recommended to use a leading backslash (even if this not required for the annotation to work correctly).

Also, the table in that documentation says that it is not recommended to specify which method is meant to be covered.

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+            $this->assertNotContains(self::FOREIGN_KEY_CHECK_STRING_START, $truncateTableSQL);
+            $this->assertNotContains(self::FOREIGN_KEY_CHECK_STRING_END, $truncateTableSQL);
+        }
+
+        $this->assertContains(self::TEST_TABLE_NAME, $truncateTableSQL);
+    }
+
+    /**
+     * @return array
```
greg0ire:
```suggestion
     * @return list<array{Driver, bool}>
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,81 @@
...
+    const FOREIGN_KEY_CHECK_STRING_END = ';SET FOREIGN_KEY_CHECKS = 1;';
+    const TEST_TABLE_NAME = 'test_table_name';
+
+    /**
+     * @param Driver $driver
+     * @param bool   $hasForeignKeyCheckString
+     *
+     * @dataProvider purgeForDifferentDriversProvider
+     */
+    public function testPurgeForDifferentDrivers(Driver $driver, $hasForeignKeyCheckString)
```
greg0ire:
```suggestion
    /**
     * @dataProvider purgeForDifferentDriversProvider
     */
    public function testPurgeForDifferentDrivers(Driver $driver, bool $hasForeignKeyCheckString): void
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,81 @@
...
+            [$this->createMock(AbstractSQLiteDriver::class), false],
+        ];
+    }
+
+    /**
+     * @param Driver $driver
+     *
+     * @return string
+     */
+    private function getTruncateTableSQLForDriver(Driver $driver)
```
greg0ire:
```suggestion
    private function getTruncateTableSQLForDriver(Driver $driver): string
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
croensch: Sorry, i forgot running the added test after rebasing. Anyway i fixed it up now.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
croensch: You mean a comment in the code or just here?

I have a test which relies on the auto-incremented ids of a small number of fixtures. I can't purge with delete because that does not reset the auto-increment to 1. I can't purge with truncate, which would reset the auto-increment, because that does not work on MySQL once i have a single relation mapped. In both cases i can only drop & create the database before loading the fixtures again.

greg0ire: A code comment could be great.

> because that does not work on MySQL once i have a single relation mapped

Why does it not work? That's what the comment should explain, possibly with references to the MySQL docs.

croensch:
I initially picked up and rebased this PR to revive _or_ end the discussion around this problem. Now that i looked into it, i understand a few things better:

- The ORMPurger touches all tables anyway, so for MySQL where the TRUNCATE is a drop/re-create internally, it should theoretically not be much faster than the current workaround of running the drop & create commands. Still, it would be nice if this feature just works on MySQL too.
- In AliceDataFixtures they implemented a fix referencing the original PR. We should not break that, i don't think it does, but it got me thinking... https://github.com/theofidry/AliceDataFixtures/blob/master/src/Bridge/Doctrine/Purger/Purger.php
- So i like how DBUnit restores the variables state, perfectly fine pattern to be least-invasive. https://github.com/sebastianbergmann/dbunit/blob/master/src/Operation/Truncate.php
- In DBUnit but also this PR i don't like how it's done around every TRUNCATE statement as it may spam the logger. I would rewrite it to call a method (like in DBUnit) before and after the tables loop.
- I still think it's a good idea to sniff the connection instead of the platform since only `AbstractMySQLDriver` is the same since 3.0 while the `AbstractMySQLPlatform` is only available since DBAL 3.3, but `3.0` is the minimum dependency.

So would you mind if i rewrite it to address all this? In the same PR?

greg0ire:
> DBAL 3.3, but 3.0 is the minimum dependency.

I think it's fine to bump the minimum dependency to 3.3, sniffing the ~connection~ driver (why did you mention the connection?) is not a great idea because of middlewares that can wrap the driver in an arbitrary number of decorators.

> So would you mind if i rewrite it to address all this? In the same PR?

Sure, go ahead :slightly_smiling_face:

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -280,4 +288,38 @@ private function getDeleteFromTableSQL(string $tableName, AbstractPlatform $plat
 
         return 'DELETE FROM ' . $tableIdentifier->getQuotedName($platform);
     }
+
+    private function disableForeignKeyChecksForMySQL(Connection $connection)
```
greg0ire:
```suggestion
    private function disableForeignKeyChecksForMySQL(Connection $connection): void
```

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -280,4 +288,38 @@ private function getDeleteFromTableSQL(string $tableName, AbstractPlatform $plat
...
+        }
+
+        // see MySQL TRUNCATE TABLE Statement: fails for an InnoDB or NDB table
+        // if there are any FOREIGN KEY constraints from other tables that
+        // reference the table.
+        $connection->executeStatement('SET @DOCTRINE_OLD_FOREIGN_KEY_CHECKS = @@FOREIGN_KEY_CHECKS');
+        $connection->executeStatement('SET FOREIGN_KEY_CHECKS = 0');
+    }
+
+    private function enableForeignKeyChecksForMySQL(Connection $connection)
```
greg0ire:
```suggestion
    private function enableForeignKeyChecksForMySQL(Connection $connection): void
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,103 @@
...
+use Doctrine\ORM\ORMSetup;
+use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Tests\Common\DataFixtures\TestEntity\Link;
+
+class ORMPurgerForeignKeyCheckTest extends BaseTest
+{
+    public const TEST_CLASS_NAME = Link::class;
+    public const TEST_TABLE_NAME = 'link';
+
+    /** @return MappingDriver */
```
greg0ire: Why not use a native return type hint here?

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,103 @@
...
+            $connection
+                ->expects($this->exactly(1))
+                ->method('executeStatement')
+                ->withConsecutive([$this->stringContains(self::TEST_TABLE_NAME)]);
+        }
+
+        $purger->purge();
+    }
+
+    /** @return list<array{AbstractPlatform, int, bool}> */
```
greg0ire:
```suggestion
    /** @return list<array{AbstractPlatform, ORMPurger::PURGE_MODE_*, bool}> */
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,103 @@
...
+                $metadata->setPrimaryTable(['name' => self::TEST_TABLE_NAME]);
+                $metadata->setIdentifier(['id']);
+            });
+        $metadataDriver->method('isTransient')->willReturn(false);
+
+        return $metadataDriver;
+    }
+
+    /** @return Connection */
+    protected function getMockConnectionForPlatform(AbstractPlatform $platform)
```
greg0ire:
```suggestion
    /** @return Connection&MockObject */
    protected function getMockConnectionForPlatform(AbstractPlatform $platform): Connection
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
croensch: I was doing it like the BaseTest::getMock*() functions. No return type, but hinting the real class.

I guess like ORMExecutorTest::getMockFixture() is the way to go?

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
croensch: Yes! Let me figure out how i can add the change to the previous function via this UI.

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
croensch: Cool, did not know.

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,103 @@
...
+use Doctrine\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Tests\Common\DataFixtures\TestEntity\Link;
+
+class ORMPurgerForeignKeyCheckTest extends BaseTest
+{
+    public const TEST_CLASS_NAME = Link::class;
+    public const TEST_TABLE_NAME = 'link';
+
+    /** @return MappingDriver */
+    protected function getMockMetadataDriver()
```
croensch:
```suggestion
    /** @return MappingDriver&MockObject */
    protected function getMockMetadataDriver(): MappingDriver
```

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -5,14 +5,18 @@
 namespace Doctrine\Common\DataFixtures\Purger;
 
 use Doctrine\Common\DataFixtures\Sorter\TopologicalSorter;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
```
croensch, line 15:
```suggestion
use PHPUnit\Framework\MockObject\MockObject;

```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
greg0ire: Yes, because `BaseTest::getMock*()` do not return _phpunit_ mocks, unlike what you are doing.

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,104 @@
...
+    public const TEST_CLASS_NAME = Link::class;
+    public const TEST_TABLE_NAME = 'link';
+
+    /** @return MappingDriver&MockObject */
+    protected function getMockMetadataDriver(): MappingDriver
+    {
+        $metadataDriver = $this->createMock(MappingDriver::class);
+        $metadataDriver->method('getAllClassNames')->willReturn([self::TEST_CLASS_NAME]);
+        $metadataDriver->method('loadMetadataForClass')
+            ->willReturnCallback(static function (string $className, ClassMetadata $metadata) {
```
greg0ire:
```suggestion
            ->willReturnCallback(static function (string $className, ClassMetadata $metadata): void {
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,104 @@
...
+                ->expects($this->exactly(1))
+                ->method('executeStatement')
+                ->withConsecutive([$this->stringContains(self::TEST_TABLE_NAME)]);
+        }
+
+        $purger->purge();
+    }
+
+    /** @return list<array{AbstractPlatform, ORMPurger::PURGE_MODE_*, bool}> */
+    public function purgeForDifferentPlatformsProvider()
```
greg0ire:
```suggestion
    public function purgeForDifferentPlatformsProvider(): array
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
croensch: Indeed, i got confused there. Solved now.

croensch:
Sorry for the spam. I got nothing to add or change - it's finished for me.

Regarding the DBAL (Common?) dependency i also check the previous class in a few easy-to-delete lines. Just in case you want to postpone upping the dependency.

greg0ire (approved):

greg0ire:
Reviewers: to be squash-merged

croensch:
@SenseException @derrabus can we get another look at this?

SenseException (approved):
Having a functional test with the foreign keys would be a nice-to-have, even though I don't expect breaking changes in the methods.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -280,4 +288,38 @@ private function getDeleteFromTableSQL(string $tableName, AbstractPlatform $plat
...
+
+        $connection->executeStatement('SET FOREIGN_KEY_CHECKS = @DOCTRINE_OLD_FOREIGN_KEY_CHECKS');
+    }
+
+    private function isMySQL(Connection $connection): bool
+    {
+        $platform = $connection->getDatabasePlatform();
+
+        if (! class_exists(AbstractMySQLPlatform::class)) {
+            // before DBAL 3.3
```
derrabus, line 319: We don't need to support ancient DBAL versions anymore.

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -163,6 +169,8 @@ public function purge()
                 $connection->executeStatement($platform->getTruncateTableSQL($tbl, true));
             }
         }
+
+        $this->enableForeignKeyChecksForMySQL($connection);
```
derrabus, line 173: If anything inside the foreach throws, we leave the connection in a state of disabled foreign key checks, don't we? We should do something about that.

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,104 @@
...
+use Doctrine\Tests\Common\DataFixtures\TestEntity\Link;
+use PHPUnit\Framework\MockObject\MockObject;
+
+class ORMPurgerForeignKeyCheckTest extends BaseTest
+{
+    public const TEST_CLASS_NAME = Link::class;
+    public const TEST_TABLE_NAME = 'link';
+
+    /** @return MappingDriver&MockObject */
+    protected function getMockMetadataDriver(): MappingDriver
```
derrabus, line 28:
```suggestion
    private function getMockMetadataDriver(): MappingDriver
```

tests/Doctrine/Tests/Common/DataFixtures/Purger/ORMPurgerForeignKeyCheckTest.php
```diff
@@ -0,0 +1,104 @@
...
+                $metadata->setPrimaryTable(['name' => self::TEST_TABLE_NAME]);
+                $metadata->setIdentifier(['id']);
+            });
+        $metadataDriver->method('isTransient')->willReturn(false);
+
+        return $metadataDriver;
+    }
+
+    /** @return Connection&MockObject */
+    protected function getMockConnectionForPlatform(AbstractPlatform $platform): Connection
```
derrabus, line 47:
```suggestion
    private function getMockConnectionForPlatform(AbstractPlatform $platform): Connection
```

derrabus (commented):
The 1.5.x branch is closed for features. We need to target 1.7.x now.

PR #411, opened 2022-10-21
Speedup ORMPurger

### LINKED ISSUE #410: Feature request: ORMPurger cache statements
I use ORMPurger on every test in `setUp` method:
1. Call `$this->getFixtureExecutor()->execute($this->getFixtureLoader()->getFixtures());`.
2. ORMExecutor internally calls `$this->purger->purge();` independently of fixtures passed
3. ORMPurger::purge do a lot of manipulations (use metadata, getSchemaAssetsFilter, getFilterSchemaAssetsExpression, etc),  and finally run sql statements.

I think it is possible to prevent extra manipulations in ORMPurger and cache, reuse sql statements. What do you think? What type of cache we can use (static variable, for example)?

dmitryuk:
Closes #410

I did a lot of improvements:
- Some variables are outside loops
- Added sql statements cache with static variable `$cachedSqlStatements`
- `array_search` replaced to `in_array`

### DIFF
--- a/src/Purger/ORMPurger.php
+++ b/src/Purger/ORMPurger.php
@@ -10,11 +10,11 @@ use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 
+use function array_map;
 use function array_reverse;
-use function array_search;
 use function assert;
 use function count;
-use function is_callable;
+use function in_array;
 use function method_exists;
 use function preg_match;
 
@@ -42,6 +42,9 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
      */
     private array $excluded;
 
+    /** @var list<string>|null */
+    private ?array $cachedSqlStatements = null;
+
     /**
      * Construct new purger instance.
      *
@@ -63,7 +66,8 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
      */
     public function setPurgeMode($mode)
     {
-        $this->purgeMode = $mode;
+        $this->purgeMode           = $mode;
+        $this->cachedSqlStatements = null;
     }
 
     /**
@@ -79,7 +83,8 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
     /** @inheritDoc */
     public function setEntityManager(EntityManagerInterface $em)
     {
-        $this->em = $em;
+        $this->em                  = $em;
+        $this->cachedSqlStatements = null;
     }
 
     /**
@@ -95,7 +100,19 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
     /** @inheritDoc */
     public function purge()
     {
-        $classes = [];
+        $connection = $this->em->getConnection();
+        array_map([$connection, 'executeStatement'], $this->getPurgeStatements());
+    }
+
+    /** @return list<string> */
+    private function getPurgeStatements(): array
+    {
+        if ($this->cachedSqlStatements !== null) {
+            return $this->cachedSqlStatements;
+        }
+
+        $connection = $this->em->getConnection();
+        $classes    = [];
 
         foreach ($this->em->getMetadataFactory()->getAllMetadata() as $metadata) {
             if ($metadata->isMappedSuperclass || (isset($metadata->isEmbeddedClass) && $metadata->isEmbeddedClass)) {
@@ -108,7 +125,7 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
         $commitOrder = $this->getCommitOrder($this->em, $classes);
 
         // Get platform parameters
-        $platform = $this->em->getConnection()->getDatabasePlatform();
+        $platform = $connection->getDatabasePlatform();
 
         // Drop association tables first
         $orderedTables = $this->getAssociationTables($commitOrder, $platform);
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks! Since this is an improvement, you should target 1.6.x

dmitryuk:
> Thanks! Since this is an improvement, you should target 1.6.x

Done

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -43,6 +43,9 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
      */
     private $excluded;
 
+    /** @var string[]|null */
+    private static $cachedSqlStatements = null;
```
stof: I would not make it a static property. We can have multiple object managers, which will need different sql statements (different entities, different platforms). A global cache would require being keyed by something distinguishing those cases

stof (changes requested):

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
dmitryuk: I removed static, because it is not required.
On project side I can define static variable with ORMPurger instance and reuse it

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -96,73 +99,80 @@ public function getObjectManager()
     /** @inheritDoc */
     public function purge()
     {
-        $classes = [];
+        $connection = $this->em->getConnection();
+
+        if ($this->cachedSqlStatements === null) {
```
stof: Calls to `setEntityManager` or `setPurgeMode` must invalidate the cache (at least if the new value is different from the old one

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -96,73 +99,80 @@ public function getObjectManager()
...
+            $this->cachedSqlStatements = [];
+            $classes                   = [];
+
+            foreach ($this->em->getMetadataFactory()->getAllMetadata() as $metadata) {
```
stof: instead of wrapping all those statements inside a if, I suggest creating `private function getPurgeStatements: array` to hold all that logic (with an early return if the cache is available)

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -43,6 +43,9 @@ class ORMPurger implements PurgerInterface, ORMPurgerInterface
      */
     private $excluded;
 
+    /** @var string[]|null */
```
stof: should be `list<string>|null` to be more precise.

dmitryuk: Done, but will be great if you describe the difference. Also it could set `array<string>`

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
dmitryuk: done

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
dmitryuk: done

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -81,6 +86,7 @@ public function getPurgeMode()
     public function setEntityManager(EntityManagerInterface $em)
     {
         $this->em = $em;
+        $this->invalidateCache();
```
stof: no need to create a private method for that IMO. You could reset the cache inline

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -96,6 +102,24 @@ public function getObjectManager()
...
+            $this->cachedSqlStatements = $this->getPurgeStatements();
+        }
+
+        array_map([$connection, 'executeStatement'], $this->cachedSqlStatements);
+    }
+
+    /**
+     * @return list<string>
+     *
+     * @throws Exception
```
stof: should be removed IMO

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -96,6 +102,24 @@ public function getObjectManager()
     /** @inheritDoc */
     public function purge()
     {
+        $connection = $this->em->getConnection();
+
+        if ($this->cachedSqlStatements === null) {
```
stof: I would put this handling of the cache in the private method instead:

```php
public function purge()
{
    array_map([$connection, 'executeStatement'], $this->getPurgeStatements());
}

private function getPurgeStatements(): array
{
    if ($this->cachedSqlStatements !== null) {
        return $this->cachedSqlStatements;
    }

    // ...

    return $this->cachedSqlStatements = $statements;
}
```

dmitryuk: done

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
dmitryuk: done

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
dmitryuk: done

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -129,40 +145,43 @@ public function purge()
...
+            $connectionConfiguration,
             'getSchemaAssetsFilter'
-        ) ? $connection->getConfiguration()->getSchemaAssetsFilter() : null;
+        ) ? $connectionConfiguration->getSchemaAssetsFilter() : null;
 
+        $this->cachedSqlStatements = [];
         foreach ($orderedTables as $tbl) {
             // If we have a filter expression, check it and skip if necessary
-            if (! $emptyFilterExpression && ! preg_match($filterExpr, $tbl)) {
+            if (! empty($filterExpr) && ! preg_match($filterExpr, $tbl)) {
```
derrabus: Why is inlining `$emptyFilterExpression` into the loop a performance improvement? `empty($filterExpr)` should be an invariant expression, why evaluate it on each iteration?

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
stof: `string[]` (or `array<string>` which is the same but with a syntax supporting union types for the type of values) is only specifying the type of values. It is technically the same than `array<int|string, string>`.

`list<string>` is a lot more strict. It forces to have a list, i.e. sequential integer keys (Psalm has full support for lists. PHPStan does not yet have it and treats `list<string>` as an alias of `array<int, string>`, so not reporting cases where keys are not sequential)

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
dmitryuk: This change is not about performance, but to fix IDE error. I don't think it can impact much
image

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
```diff
@@ -96,7 +101,18 @@ public function getObjectManager()
     /** @inheritDoc */
     public function purge()
     {
-        $classes = [];
+        $connection = $this->em->getConnection();
+        array_map([$connection, 'executeStatement'], $this->getPurgeStatements());
+    }
+
+    private function getPurgeStatements(): array
```
greg0ire:
```suggestion
    /** @return list<string> */
    private function getPurgeStatements(): array
```

lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php
derrabus: In that case, I suggest to revert it.

dmitryuk: Done, but I reversed a condition from `!$emptyFilterExpression` to `$hasFilterExpression` - easier to read

dmitryuk:
Hello. What next?

dmitryuk:
Rebased 1.6.x, please approve workflow again

dmitryuk:
@stof please check PR again

dmitryuk:
Rebased again, still waiting for merge..

greg0ire (approved):

derrabus (approved):

derrabus:
Thank you @dmitryuk!

PR #413, opened 2022-12-22
Fix reference update

VincentLanglet:
Related to https://github.com/doctrine/data-fixtures/pull/409 (cc @greg0ire)
 
When the manager does not contains the reference of an object and need to update it, 
`$this->references[$name]` was updated but not `$this->referencesByClass[$class][$name]`.

This PR fix this.


greg0ire (approved):

greg0ire:
Thanks @VincentLanglet!

PR #389, opened 2022-04-21
Improve phpdoc types

BafS:
- Removed some unhelpful comment (constructors)
- Added some more specific types in the phpdoc

PR #416, opened 2023-01-04
Use a more accurate type 

### LINKED ISSUE #414: ORMExecutor->execute expects diffrented  class than Loader->getFixures() returns
I get the following error from latest versions:
```
  Line   tests/Functional/FunctionalTestCase.php                                                                                                                      
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  91     Parameter #1 $fixtures of method Doctrine\Common\DataFixtures\Executor\ORMExecutor::execute() expects array<Doctrine\Common\DataFixtures\FixtureInterface>,  
         array<class-string<Doctrine\Common\DataFixtures\OrderedFixtureInterface>|int, Doctrine\Common\DataFixtures\OrderedFixtureInterface> given.    
```

I think it could be resolved if `OrderedFixtureInterface` extends from `FixtureInterface`

greg0ire:
`getFixtures()` doesn't always return instances of `OrderedFixtureInterface`
(for instance if you don't use those at all).

Fixes #414 , and should also fix the build of the bundle (see https://github.com/doctrine/DoctrineFixturesBundle/pull/375 for an example of a broken build)

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/Loader.php
+++ b/lib/Doctrine/Common/DataFixtures/Loader.php
@@ -47,7 +47,7 @@ class Loader
     /**
      * Array of ordered fixture object instances.
      *
-     * @psalm-var array<class-string<OrderedFixtureInterface>, OrderedFixtureInterface>|list<OrderedFixtureInterface>
+     * @psalm-var array<class-string<FixtureInterface>|int, FixtureInterface>
      */
     private $orderedFixtures = [];
 
@@ -181,7 +181,7 @@ class Loader
     /**
      * Returns the array of data fixtures to execute.
      *
-     * @psalm-return array<class-string<OrderedFixtureInterface>|int, OrderedFixtureInterface>
+     * @psalm-return array<class-string<FixtureInterface>|int, FixtureInterface>
      */
     public function getFixtures()
     {
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26,11 +26,6 @@ parameters:
 			path: lib/Doctrine/Common/DataFixtures/Loader.php
 
 		-
-			message: "#^Property Doctrine\\\\Common\\\\DataFixtures\\\\Loader\\:\\:\\$orderedFixtures \\(array\\<class\\-string\\<Doctrine\\\\Common\\\\DataFixtures\\\\OrderedFixtureInterface\\>\\|int, Doctrine\\\\Common\\\\DataFixtures\\\\OrderedFixtureInterface\\>\\) does not accept array\\<class\\-string\\<Doctrine\\\\Common\\\\DataFixtures\\\\FixtureInterface\\>, Doctrine\\\\Common\\\\DataFixtures\\\\FixtureInterface\\>\\.$#"
-			count: 2
-			path: lib/Doctrine/Common/DataFixtures/Loader.php
-
-		-
 			message: "#^Call to an undefined method Doctrine\\\\Persistence\\\\ObjectManager\\:\\:getReference\\(\\)\\.$#"
 			count: 2
 			path: lib/Doctrine/Common/DataFixtures/ProxyReferenceRepository.php


lib/Doctrine/Common/DataFixtures/Loader.php
```diff
@@ -181,7 +183,7 @@ public function addFixture(FixtureInterface $fixture)
     /**
      * Returns the array of data fixtures to execute.
      *
-     * @psalm-return array<class-string<OrderedFixtureInterface>|int, OrderedFixtureInterface>
+     * @psalm-return array<class-string<OrderedFixtureInterface&FixtureInterface>|int, OrderedFixtureInterface&FixtureInterface>
```
ghost:
```suggestion
     * @psalm-return array<class-string<OrderedFixtureInterface|FixtureInterface>|int, OrderedFixtureInterface|FixtureInterface>
```

Is it not `|` also this still wont fix the phpstan error but that should be done in major release right as that would be a breaking change?

greg0ire: > Is it not `|`

No? Why on earth would it? It's not an `OrderedFixtureInterface` OR a `FixtureInterface`, it is both, right?


>  this still wont fix the phpstan error

why not? have you tried it?

ghost: > > Is it not `|`
> 
> No? Why on earth would it? It's not an `OrderedFixtureInterface` OR a `FixtureInterface`, it is both, right?

I think your right because `OrderedFixtureInterface` and `FixtureInterface` would both be implemented
> 
> > this still wont fix the phpstan error
> 
> why not? have you tried it?

No but i see you update the phpstan ignore error so i assumed its not fixed, but dont know what we should change without breaking to fix it

greg0ire: > > have you tried it?

> No 

Well what are you waiting for?

ghost: > > > have you tried it?
> 
> > No
> 
> Well what are you waiting for?

I confirmed it works for my case, but i still have the question why we need to keep these ignored https://github.com/doctrine/data-fixtures/pull/416/files#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efaR29?

greg0ire: Fair question, let me take a closer look.

greg0ire: :thinking: I know think what has been done in https://github.com/doctrine/data-fixtures/pull/389/files#diff-7d95dacf620e8b57cfdb926d6f6cc52dbf52687cf3095881544bf9c6e3abc4d8 was not only imprecise, but also plain wrong: `getFixtures` may return fixtures that are not ordered fixtures! The property is called `orderedFixtures` can contain fixtured ordered because the fixtures implement `OrderedFixtureInterface`, or because they implement `DependentFixtureInterface`, and if they don't, the order will be the order in which they were added. Thanks for making me take a closer look, I'll change this PR!

greg0ire:
Cc @BafS

ghost (approved):

lib/Doctrine/Common/DataFixtures/Loader.php
```diff
@@ -47,7 +47,7 @@ class Loader
     /**
      * Array of ordered fixture object instances.
      *
-     * @psalm-var array<class-string<OrderedFixtureInterface>, OrderedFixtureInterface>|list<OrderedFixtureInterface>
+     * @psalm-var array<class-string<FixtureInterface>, FixtureInterface>|list<FixtureInterface>
```
franmomu: I think, for consistency, `$orderedFixtures` property and `getLoader()` method should have the same type (in this case maybe `int` as a key makes more sense than a `list` meaning that it is an ordered array).

stof (approved):

PR #419, opened 2023-01-05
Update ReferenceRepository::getRealClass() to match with doctrine\common

### LINKED ISSUE #417: Cant use an entity with ReferenceRepository if its name ends with Proxy
Since v1.6.0 our tests fail when loading fixtures for entities with a Proxy suffix in their name.

For example with the entity `Path\To\Entities\EntityProxy` we get a `Doctrine\Persistence\Mapping\MappingException: Class 'Path\To\Entities\Entity' does not exist` error.

I think the bug was introduced in https://github.com/doctrine/data-fixtures/pull/409. The `getRealClass` removes the `Proxy` suffix. Maybe the check to remove the suffix `Proxy` could be made more specific to actual proxies?

mdevlamynck:
Fixes #417.

This fixes ReferenceRepository::getRealClass() to match with the behaviour of the equiavlent function in doctrine\common ClassUtils::getRealClass().

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
@@ -8,6 +8,7 @@ use BadMethodCallException;
 use Doctrine\Deprecations\Deprecation;
 use Doctrine\ODM\PHPCR\DocumentManager as PhpcrDocumentManager;
 use Doctrine\Persistence\ObjectManager;
+use Doctrine\Persistence\Proxy;
 use OutOfBoundsException;
 
 use function array_key_exists;
@@ -15,6 +16,7 @@ use function array_keys;
 use function get_class;
 use function method_exists;
 use function sprintf;
+use function strrpos;
 use function substr;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -385,11 +386,13 @@ public function getManager()
      */
     protected function getRealClass($className)
     {
-        if (substr($className, -5) === 'Proxy') {
-            return substr($className, 0, -5);
+        $pos = strrpos($className, '\\' . Proxy::MARKER . '\\');
```
greg0ire, line 390: You can run `vendor/bin/phpbcf` to fix this one

greg0ire:
Can you please add a unit test for this?

stof:
As this is a naming convention coming from doctrine/persistence, I actually suggest adding an utility in doctrine/persistence (meant to replace the one from doctrine/common) instead of duplicating the logic in each package wanting to remove a dependency on doctrine/common.

mdevlamynck:
So @stof do you want me to create a separate PR to doctrine/persistence to add that utility then update my this PR?

VincentLanglet:
> As this is a naming convention coming from doctrine/persistence, I actually suggest adding an utility in doctrine/persistence (meant to replace the one from doctrine/common) instead of duplicating the logic in each package wanting to remove a dependency on doctrine/common.

That's a good idea, but I'm not sure it will help here since this library is supporting
```
"doctrine/persistence": "^1.3.3|^2.0|^3.0"
```
or it would require to drop both 1 and 2 support.

greg0ire:
>  it would require to drop both 1 and 2 support.

Seems fine given the version constraints of v3, as well as the install statistics: https://packagist.org/packages/doctrine/persistence/stats

VincentLanglet:
For moving ClassUtils to doctrine/persistence I opened https://github.com/doctrine/persistence/pull/326
We will need this to be merged/released first then.

PR #421, opened 2023-01-07
Rely on object manager proxy resolution

### LINKED ISSUE #417: Cant use an entity with ReferenceRepository if its name ends with Proxy

VincentLanglet:
Suggested by https://github.com/doctrine/persistence/pull/326#issuecomment-1374466942 cc @greg0ire 

Closes https://github.com/doctrine/data-fixtures/pull/419
Closes https://github.com/doctrine/data-fixtures/issues/417

This would be great to try it in a real project.

### DIFF
--- a/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
+++ b/lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
@@ -15,7 +15,6 @@ use function array_keys;
 use function get_class;
 use function method_exists;
 use function sprintf;
-use function substr;
 
 /**
  * ReferenceRepository class manages references for
@@ -82,7 +81,7 @@ class ReferenceRepository
     {
         // In case Reference is not yet managed in UnitOfWork
         if (! $this->hasIdentifier($reference)) {
-            $class = $this->manager->getClassMetadata($this->getRealClass(get_class($reference)));
+            $class = $this->manager->getClassMetadata(get_class($reference));
 
             return $class->getIdentifierValues($reference);
         }
@@ -385,11 +384,7 @@ class ReferenceRepository
      */
     protected function getRealClass($className)
     {
-        if (substr($className, -5) === 'Proxy') {
-            return substr($className, 0, -5);
-        }
-
-        return $className;
+        return $this->manager->getClassMetadata($className)->getName();
     }
 
     /**
--- a/tests/Doctrine/Tests/Common/DataFixtures/ReferenceRepositoryTest.php
+++ b/tests/Doctrine/Tests/Common/DataFixtures/ReferenceRepositoryTest.php
@@ -14,7 +14,6 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Proxy;
 use Doctrine\Tests\Common\DataFixtures\TestEntity\Role;
 use OutOfBoundsException;
-use stdClass;
 
 class ReferenceRepositoryTest extends BaseTest
 {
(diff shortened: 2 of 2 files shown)


greg0ire:
You can use the `Closes` keyword for PRs that should be closed if we merge this as well I think. Needs `vendor/bin/phpcbf`

greg0ire:
Please improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).

This is hard to understand, so a good commit message is in order. Here, what happens is that several proxy systems can be in use (the one from `doctrine/common`, and maybe the one Nicolas Grekas recently introduced). By using delegating to the object manager layer, we don't need to know.

greg0ire:
> This would be great to try it in a real project.

@mdevlamynck, can you maybe help with this?

greg0ire (approved):

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -385,11 +384,7 @@ public function getManager()
      */
     protected function getRealClass($className)
     {
-        if (substr($className, -5) === 'Proxy') {
-            return substr($className, 0, -5);
-        }
-
-        return $className;
+        return $this->manager->getMetadataFactory()->getMetadataFor($className)->getName();
```
franmomu:
```suggestion
        return $this->manager->getClassMetadata($className)->getName();
```
This would allow to simplify tests I think

mdevlamynck:
@greg0ire Works in our project.

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
```diff
@@ -82,7 +81,7 @@ protected function getIdentifier($reference, $uow)
     {
         // In case Reference is not yet managed in UnitOfWork
         if (! $this->hasIdentifier($reference)) {
-            $class = $this->manager->getClassMetadata($this->getRealClass(get_class($reference)));
+            $class = $this->manager->getClassMetadata(get_class($reference));
```
VincentLanglet, line 84: No need to use `getRealClass` here since the getClassMetada is already doing the job.

lib/Doctrine/Common/DataFixtures/ReferenceRepository.php
VincentLanglet: Thanks, done.

greg0ire:
Might help:

```
cat .git/hooks/pre-push
#!/bin/bash
set -e
echo ''|vendor/bin/phpcs
vendor/bin/phpstan
vendor/bin/psalm
vendor/bin/phpunit
```

VincentLanglet:
Cs is fixed now

greg0ire (approved):

tests/Doctrine/Tests/Common/DataFixtures/ReferenceRepositoryTest.php
```diff
@@ -129,15 +128,21 @@ public function testUndefinedReference(): void
...
     public function testThrowsExceptionAddingDuplicatedReference(): void
     {
-        $referenceRepository = new ReferenceRepository($this->getMockSqliteEntityManager());
-        $referenceRepository->addReference('duplicated_reference', new stdClass());
+        $em                  = $this->getMockSqliteEntityManager();
+        $referenceRepository = new ReferenceRepository($em);
+
+        $schemaTool = new SchemaTool($em);
+        $schemaTool->dropSchema([]);
+        $schemaTool->createSchema([$em->getClassMetadata(Role::class)]);
```
franmomu: Are these lines necessary? (since apparently there are no calls to `persist` or `flush`)

VincentLanglet: Useless indeed so I removed them.

greg0ire: No need to make a new commit, please amend the first one.

greg0ire (approved):

franmomu (approved):

greg0ire:
Thanks @VincentLanglet !

PR #425, opened 2023-01-30
Document possibly null property

### LINKED ISSUE #424: Psalm error for referenceRepository
i followed the example given [here](https://symfony.com/bundles/DoctrineFixturesBundle/current/index.html#accessing-services-from-the-fixtures) to inject a service into the fixture.

now when i'm running psalm it prints out a warning that `referenceRepository` is not set in the constructor

```
ERROR: PropertyNotSetInConstructor - src/DataFixtures/UserFixtures.php:12:7 - Property App\DataFixtures\UserFixtures::$referenceRepository is not defined in constructor of App\DataFixtures\UserFixtures or in any methods called in the constructor (see https://psalm.dev/074)
class UserFixtures extends Fixture
```

my assumption is that it is because in `AbstractFixture` the type hint for `referenceRepository` only mentions `ReferenceRepository` but not null which could theoretically happen

greg0ire:
Fixes #424

### DIFF
--- a/src/AbstractFixture.php
+++ b/src/AbstractFixture.php
@@ -7,6 +7,8 @@ namespace Doctrine\Common\DataFixtures;
 use BadMethodCallException;
 use Doctrine\Deprecations\Deprecation;
 
+use function assert;
+
 /**
  * Abstract Fixture class helps to manage references
  * between fixture classes in order to set relations
@@ -17,7 +19,7 @@ abstract class AbstractFixture implements SharedFixtureInterface
     /**
      * Fixture reference repository
      *
-     * @var ReferenceRepository
+     * @var ReferenceRepository|null
      */
     protected $referenceRepository;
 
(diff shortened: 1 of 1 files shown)


src/AbstractFixture.php
```diff
@@ -29,6 +31,13 @@ public function setReferenceRepository(ReferenceRepository $referenceRepository)
         $this->referenceRepository = $referenceRepository;
     }
 
+    private function referenceRepository(): ReferenceRepository
```
derrabus:
```suggestion
    private function getReferenceRepository(): ReferenceRepository
```

src/AbstractFixture.php
```diff
@@ -17,9 +19,9 @@ abstract class AbstractFixture implements SharedFixtureInterface
     /**
      * Fixture reference repository
      *
-     * @var ReferenceRepository
+     * @var ReferenceRepository|null
      */
-    protected $referenceRepository;
+    protected $referenceRepository = null;
```
derrabus: Untyped properties are `null` by default and don't need a `null` initialization. The docblock change should be enough.

derrabus (changes requested):

derrabus (approved):

PR #428, opened 2023-02-17
Add blank line after code block start

greg0ire:
Not having it makes the whole block invalid.

Source: https://github.com/doctrine/doctrine-website/pull/476#issuecomment-1434252551


linawolf (approved):

stof:
Maybe the CI should have a workflow trying to build the documentation when there are changes in the `docs` folder, so that mistakes like that are caught in the CI and not just when building the website after the merge (where contributors would not see it)

linawolf:
Should be easy to do after the update of the rst-parser

PR #436, opened 2023-04-03
Maintenance

greg0ire:


greg0ire:
:thinking: I'm wondering if I need to install the mongodb extension to fix PHPStan, cc @alcaeus @franmomu

EDIT: doing so locally doesn't fix the issue.

franmomu:
> 🤔 I'm wondering if I need to install the mongodb extension to fix PHPStan, cc @alcaeus @franmomu
> 
> EDIT: doing so locally doesn't fix the issue.

could be something related to the PHP version set in 

https://github.com/doctrine/data-fixtures/blob/4892c74dd04a3ec7404b9586272d722325905c58/phpstan.neon.dist#L2

which is different from the PHP version running the job? 

`doctrine/mongodb-odm` 2.5 only supports PHP >= 7.4

derrabus:
Indeed. In other Doctrine projects, we've changed this setting to the highest supported PHP version. We should do the same here.

greg0ire:
Ah well spotted @franmomu , that does fix the issue. I wouldn't have thought of checking this, thanks a lot!

phpunit.xml.dist
```diff
@@ -12,9 +12,9 @@
         </testsuite>
     </testsuites>
 
-    <filter>
-        <whitelist>
+    <coverage>
```
greg0ire, line 15: I guess this means we only get coverage for PHP >= 7.3

derrabus: Probably. FWIW, [nobody](https://packagist.org/packages/doctrine/data-fixtures/php-stats#1.6)'s using this package on PHP 7.2/7.3 anymore anyway.

greg0ire: Ok then it's fine, and we should drop support for those in the next minor then.

derrabus (approved):

PR #437, opened 2023-04-03
doctrine/deprecations is a mandatory dependency

derrabus:
We trigger deprecations from non-test classes and need the Doctrine deprecations library for that. Because of that, it has to be a mandatory dependency.

This change probably doesn't make that much of a difference because data fixtures is most likely used in projects that depend on other libraries that pull the `doctrine/deprecations` dependency already, like DBAL or ORM.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
-        "doctrine/persistence": "^1.3.3|^2.0|^3.0"
+        "doctrine/deprecations": "^0.5.3 || ^1.0",
+        "doctrine/persistence": "^1.3.3 || ^2.0 || ^3.0"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -15,7 +15,8 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
-        "doctrine/persistence": "^1.3.3|^2.0|^3.0"
+        "doctrine/deprecations": "^v0.5.3 || ^1.0",
```
greg0ire:
```suggestion
        "doctrine/deprecations": "^0.5.3 || ^1.0",
```

derrabus: Oops, where did that come from? 🤔

greg0ire (approved):

PR #445, opened 2023-07-29
Add static analysis and cs config files to .gitattributes

VincentLanglet:
Hi @greg0ire 

Those 5 files can also be added to the gitattributes one

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,8 @@
 /.github export-ignore
 /.gitignore export-ignore
 /phpunit.xml.dist export-ignore
+/phpcs.xml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpstan-baseline.neon export-ignore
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

greg0ire:
Thanks! This makes me wish that alternate syntax I found a few years back where you state what should be included instead of what shouldn't was understood by all tools (Satis does not IIRC)

VincentLanglet:
> Thanks! This makes me wish that alternate syntax I found a few years back where you state what should be included instead of what shouldn't was understood by all tools (Satis does not IIRC)

Another solution would be to have something like sonata-devkit to share files like `gitattributes` ^^'
https://github.com/sonata-project/dev-kit/blob/master/templates/project/.gitattributes.twig

greg0ire:
Yeah… I'm afraid we're not there yet :P

PR #444, opened 2023-06-17
Replace usage of deprecated LifecycleEventArgs method

acrobat:
Fixes #443

PR #449, opened 2023-08-17
Replace call to deprecated `LifecycleEventArgs::getEntityManager()`

mbabker:
Similar to #444 but now replacing `getEntityManager()` with `getObjectManager()`.


greg0ire (approved):

greg0ire:
Thanks @mbabker !

PR #171, opened 2015-02-26
Fix purging non-public schema tables

molchanoviv:
Hello. This PR fixes purging non-default schema tables in PostgreSQL.

PR #452, opened 2023-10-29
Support DBAL 4 and ORM 3

derrabus:


derrabus:
The test failure looks like a bug in ORM or DBAL.

tests/Common/DataFixtures/TestEntity/User.php
```diff
@@ -43,13 +51,18 @@ class User
      *
      * @var Collection<int, User>
      */
+    #[ORM\ManyToMany(targetEntity: self::class, inversedBy: 'authors')]
+    #[ORM\JoinTable(name: 'author_reader', schema: 'readers')]
```
greg0ire, line 55: Is SQLite supposed to support this (The `schema` argument)? I took a look at the test failure, then at that `JoinTable`, that seems related to the failure.

greg0ire: I reproduce the issue when using ORM 3 and DBAL 3, so it's not related to DBAL.

greg0ire: Ah maybe it has to do with the `RemoveNamespacedAsset` thingy.

greg0ire: ORM stopped using it in https://github.com/doctrine/orm/pull/11001

derrabus: > Is SQLite supposed to support this (The `schema` argument)?

Yes, SQLite supports schemas. You basically attach a second SQLite file and assign a schema name to it.

greg0ire: Ok, but apparently, ORM 2 does not care whether it supports schemas, but whether it can emulate them: https://github.com/doctrine/orm/blob/1a4fe6e0bb67762370937a7e6cee3da40a9122d1/lib/Doctrine/ORM/Tools/SchemaTool.php#L415

In the case of SQLite, that's dynamic: https://github.com/doctrine/dbal/blob/0ac3c270590e54910715e9a1a044cc368df282b2/src/Platforms/SqlitePlatform.php#L896-L905

When running the tests on 1.7.x, it returns `false`, which I think explains why this worked before, but does not work now.

When I say it "worked", I mean it is possible to create the schema, because as soon as you try to create a user and add it as a reader of another user, it falls apart: `Unable to prepare statement: 1, no such table: readers.author_reader`.

franmomu: I tried to reproduce this in https://github.com/doctrine/orm/pull/11079, if that's fine I guess we can maybe remove the `schema` part here since doesn't seem relevant.

franmomu: I found https://github.com/doctrine/orm/pull/9259, so apparently they won't be supported in 3.0

greg0ire: Schema emulation is not supported be supported in 3.0. But @derrabus says SQLite supports schemas without any emulation. And as I showed, data-fixtures is not enabling schema emulation anyway. I think the right course of action would be to determine what exactly we are trying to test here… why is there a `schema` argument in the first place?

greg0ire: The git pickaxe says the test was introduced in 70c915d7a6566dfe67838ffeded8b8d654a50dec
That commit introduces another test than the one that fails. It reuses the same fixture… not great.

greg0ire: After much digging, I found the associated PR, which uses another commit hash, probably because of squashing: https://github.com/doctrine/data-fixtures/pull/171

So the goal is to fix purging for non-public schema tables. I think this vocabulary means the issue is a PostgreSQL one, but the test was reworked so that it could be tested with SQLite's schema emulation by… me, in #333, following this fix: https://github.com/doctrine/orm/pull/7079

The test worked before the fix because it's a unit test: no tables were created.

> When running the tests on 1.7.x, it returns false, which I think explains why this worked before, but does not work now.

On my work laptop it returns `true` today for some reason :upside_down_face:

greg0ire: I see 2 possible ways out of this now:
1. we create seperate fixtures for separate tests.
2. we think it's cool to test the creation of the n-m table in a separate schema and attach another database as alluded to by @derrabus . That being said, I don't know if it will be that simple, because if it were, wouldn't the error message be about an unknown database? Right now the message says that there is a syntax error while creating… an index: `CREATE INDEX IDX_3155D2E5F675F31B ON readers.author_reader (author_id)`… :thinking: why is there nothing about the table?

UPD: according to [the docs](https://www.sqlite.org/lang_createindex.html), it should probably be `CREATE readers.INDEX IDX_3155D2E5F675F31B ON author_reader (author_id)`… so there might be a bug in DBAL after all.

greg0ire: Created [a PR with failing tests](https://github.com/doctrine/dbal/pull/6226), might write a fix later on, unless one of you beats me to it (feel free to push to my branch if you have the necessary permissions).

greg0ire: It looks really wrong. I think the real issue might be on how the ORM calls the DBAL.

UPD: can't be, my tests don't involve the ORM.

greg0ire: @derrabus green :)

derrabus:
Thank you for unblocking this PR, @greg0ire! _Someone_ needs to review and approve it then. 🥺

greg0ire:
I was waiting for the removal of the "Draft" status, in case you wanted to add more things, but I'll review it right now :)

greg0ire (approved):

PR #455, opened 2023-12-14
Add parameter types

GromNaN:
In `DoctrineMongoDBBundle`, we can't add parameter types to all the classes because they are not defined in the base class or the interface. https://github.com/doctrine/DoctrineMongoDBBundle/pull/812/files#diff-1571e1997e046b65a3fdffae3826e46e2a5a89adcce01ff580557d7eedc536e1R71

Adding parameter types to all the methods should not be an issues for classes that extends them. But it might for interfaces.

Also upgrading to PHP 8.1 is required for some `mixed` and union types.

### DIFF
--- a/src/AbstractFixture.php
+++ b/src/AbstractFixture.php
@@ -49,7 +49,7 @@ abstract class AbstractFixture implements SharedFixtureInterface
      *
      * @return void
      */
-    public function setReference(string $name, $object)
+    public function setReference(string $name, object $object)
     {
         $this->getReferenceRepository()->setReference($name, $object);
     }
@@ -106,7 +106,7 @@ abstract class AbstractFixture implements SharedFixtureInterface
      *
      * @see Doctrine\Common\DataFixtures\ReferenceRepository::hasReference
      *
-     * @psalm-param class-string $class
+     * @psalm-param class-string|null $class
      *
      * @return bool
      */
--- a/src/AbstractFixture.php
+++ b/src/AbstractFixture.php
@@ -45,12 +45,11 @@ abstract class AbstractFixture implements SharedFixtureInterface
      *
      * @see Doctrine\Common\DataFixtures\ReferenceRepository::setReference
      *
-     * @param string $name
      * @param object $object - managed object
      *
      * @return void
      */
-    public function setReference($name, $object)
+    public function setReference(string $name, $object)
     {
         $this->getReferenceRepository()->setReference($name, $object);
     }
@@ -63,14 +62,13 @@ abstract class AbstractFixture implements SharedFixtureInterface
      *
      * @see Doctrine\Common\DataFixtures\ReferenceRepository::addReference
      *
-     * @param string $name
      * @param object $object - managed object
      *
      * @return void
      *
      * @throws BadMethodCallException - if repository already has a reference by $name.
      */
-    public function addReference($name, $object)
+    public function addReference(string $name, object $object)
     {
         $this->getReferenceRepository()->addReference($name, $object);
     }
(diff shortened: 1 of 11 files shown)


derrabus:
This PR does too many things. Let's start small please.

Let's start with parameter types only. Bumping to 8.1 does not yet feel necessary to me. We should not have many union types in parameters and `mixed` is the only parameter type that you may add downstream.

GromNaN:
Indeed, I went too far. Reworking.

GromNaN:
It did not change the type for `ORMPurger::getJoinTableName` which requires an union ``.
Also `ReferenceRepository::setReferenceIdentity` has a `mixed` argument.

Nothing needs to be updated in tests.

src/AbstractFixture.php
```diff
@@ -45,12 +45,11 @@ private function getReferenceRepository(): ReferenceRepository
      *
      * @see Doctrine\Common\DataFixtures\ReferenceRepository::setReference
      *
-     * @param string $name
      * @param object $object - managed object
      *
      * @return void
      */
-    public function setReference($name, $object)
+    public function setReference(string $name, $object)
```
derrabus:
```suggestion
    public function setReference(string $name, object $object)
```

src/AbstractFixture.php
```diff
@@ -109,12 +106,11 @@ public function getReference($name, ?string $class = null)
      *
      * @see Doctrine\Common\DataFixtures\ReferenceRepository::hasReference
      *
-     * @param string $name
      * @psalm-param class-string $class
```
derrabus:
```suggestion
     * @psalm-param class-string|null $class
```

src/Executor/AbstractExecutor.php
```diff
@@ -76,23 +76,19 @@ public function getPurger()
     /**
      * Set the logger callable to execute with the log() method.
      *
-     * @param callable $logger
-     *
      * @return void
      */
-    public function setLogger($logger)
+    public function setLogger(callable $logger)
```
derrabus, line 81: Why is the logger a `callable` anyway? 🙈

derrabus (commented):
Looks good so far. Once this PR is merged, we might open a 2.0 branch where we bump to 8.1 and apply return types, WDYT?

src/Executor/AbstractExecutor.php
GromNaN: Line 94. 
```php
        $logger = $this->logger;
        $logger($message);
```

You think we should change it to accept a `Psr\Log\LoggerInterface`?

GromNaN:
> Once this PR is merged, we might open a 2.0 branch where we bump to 8.1 and apply return types, WDYT?

Sounds like a good idea.
There is little deprecated things to remove, and we could add return types.

src/Executor/AbstractExecutor.php
derrabus: We should look into that, yes. But probably in a separate PR.

derrabus (approved):

greg0ire:
Thanks @GromNaN !

src/Executor/AbstractExecutor.php
GromNaN: I'm not going to do that, because it's used for console output directly.

https://github.com/doctrine/DoctrineFixturesBundle/blob/c808a0c85c38c8ee265cc8405b456c1d2b38567d/Command/LoadDataFixturesDoctrineCommand.php#L137-L139

derrabus: > it's used for console output directly.

We could try to leverage `Symfony\Component\Console\Logger\ConsoleLogger`. My problem is that a closure is a bad logging abstraction. While it works for the way the bundle uses it, it sucks if you want the logs to go anywhere else, e.g. you want to use a logging framework instead.

Deprecation: #462

PR #456, opened 2023-12-14
Reference classes using current namespace in `@see` phpdoc

GromNaN:
Make `@see` references work (in PHPStorm at least).


greg0ire (approved):

greg0ire:
I think I would have picked 1.7.x but I'm fine with 1.8.x

GromNaN:
Switched to 1.7.x, doesn't matter.

greg0ire:
Thanks @GromNaN !

PR #290, opened 2018-03-20
Version 2: Make executors and purgers final

dbu:
This is a follow up to #289 (and will show the changes from #289 until it is merged) to make purgers and executors final and add return type and method type declarations where possible. i also remove pointless phpdoc comments.

This bumps the version to 2 - i suggest to keep this open and wait with merging until you feel ready to start the 2.0 version.

PR #457, opened 2023-12-14
Make Executor and Purger classes final

GromNaN:
By moving this classes to a soft `@final` in 1.8, we can make them hard `final` in 2.0 and add types to the properties.

We don't expect any further implementations to be created out of this package.

Related to this outdated PR preparing a v2 #290

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,11 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 1.8
+
+Executor and Purger classes are final, they cannot be extended.
+`AbstractExecutor` is internal. It cannot be extended or used as typehint.
+
 # Upgrade to 1.6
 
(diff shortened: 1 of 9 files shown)


UPGRADE.md
```diff
@@ -6,6 +6,10 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 1.8
+
+Executor and Purger classes are now final. You should not extend them.
```
greg0ire: You made the `AbstractExecutor` internal, should that be mentioned as well? What's the consequence of marking a class as `internal` for the users? I think this means that:
- they should not extend it
- they cannot even type hint against it in their code

GromNaN: Exact. I updated the UPGRADE doc.

greg0ire: Great! Now you just have to push it to github :stuck_out_tongue:

GromNaN: 🤦

greg0ire (approved):

SenseException (approved):

PR #459, opened 2023-12-15
Add return types on public classes and interfaces

GromNaN:
Add return types to methods from the public API.

Follow-up https://github.com/doctrine/data-fixtures/pull/455#pullrequestreview-1782094062

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -56,11 +56,6 @@
         <exclude-pattern>src/*</exclude-pattern>
     </rule>
 
-    <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint">
-        <!-- Adding return types is a BC break in most cases. -->
-        <exclude-pattern>src/*</exclude-pattern>
-    </rule>
-
     <rule ref="Squiz.Classes.ClassFileName.NoMatch">
         <exclude-pattern>tests/*</exclude-pattern>
     </rule>
--- a/src/FixtureInterface.php
+++ b/src/FixtureInterface.php
@@ -14,5 +14,5 @@ interface FixtureInterface
     /**
      * Load data fixtures with the passed EntityManager
      */
-    public function load(ObjectManager $manager);
+    public function load(ObjectManager $manager): void;
 }
--- a/src/ReferenceRepository.php
+++ b/src/ReferenceRepository.php
@@ -56,7 +56,7 @@ class ReferenceRepository
      *
      * @return array
      */
-    protected function getIdentifier(object $reference, object $uow)
+    protected function getIdentifier(object $reference, object $uow): array
     {
         // In case Reference is not yet managed in UnitOfWork
         if (! $this->hasIdentifier($reference)) {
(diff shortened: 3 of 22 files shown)


GromNaN:
I'll check why `ObjectManager::getClassMetadata()` return `null` for `Doctrine\Tests\Common\DataFixtures\TestEntity\Role`.

tests/Common/DataFixtures/Executor/ORMExecutorTest.php
```diff
@@ -52,8 +52,7 @@ public function testExecuteWithPurge(): void
         $em     = $this->getMockSqliteEntityManager();
         $purger = $this->getMockPurger();
         $purger->expects($this->once())
-            ->method('purge')
-            ->willReturn(null);
+            ->method('purge');
```
GromNaN: Updated because `purge()` return type is `void`.

GromNaN:
Psalm baseline updated. I was confused by the fact that both psalm and phpstan are used.

greg0ire (approved):

src/ReferenceRepository.php
```diff
@@ -75,7 +75,7 @@ public function __construct(ObjectManager $manager)
      *
      * @return array
```
derrabus: Remove.

tests/Common/DataFixtures/Executor/ORMExecutorTest.php
derrabus: We should backport that change to 1.8.x.

tests/Common/DataFixtures/ReferenceRepositoryTest.php
```diff
@@ -198,6 +198,9 @@ public function testGetIdentifierWhenHasNotBeenManagedYetByUnitOfWork(): void
             ->method('getIdentifierValues')
             ->with($role)
             ->willReturn($identitiesExpected);
+        $classMetadata->expects($this->once())
+            ->method('getName')
+            ->willReturn(Role::class);
```
derrabus: Why did this change become necessary? Should we backport?

GromNaN: Yes: https://github.com/doctrine/data-fixtures/pull/463

derrabus:
Rebase needed.

src/FixtureInterface.php
```diff
@@ -14,5 +14,5 @@ interface FixtureInterface
     /**
      * Load data fixtures with the passed EntityManager
      */
-    public function load(ObjectManager $manager);
+    public function load(ObjectManager $manager): void;
```
GromNaN, line 17: This change can impact every project using data-fixtures. Adding the `@return` type  annotation in the previous version will ease migration: #472

derrabus:
Can we enable this rule now?

https://github.com/doctrine/data-fixtures/blob/36d20789751e0fdef7540336d0fc0a69d6e07fa1/phpcs.xml.dist#L59-L62

phpcs.xml.dist
```diff
@@ -56,11 +56,6 @@
         <exclude-pattern>src/*</exclude-pattern>
     </rule>
 
-    <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint">
```
GromNaN, line 59: Response to https://github.com/doctrine/data-fixtures/pull/459#issuecomment-1991384911: This rule is applied by default. This config was to exclude `src` dir.

derrabus: Right. I missed that you've changed the config already. 🙈

derrabus (approved):

PR #462, opened 2023-12-18
Deprecate callable loggers

derrabus:
We should not invent our own contract for logging. There's PSR-3 for that.

This PR proposes to deprecate passing a callable to `AbstractExecutor::setLogger()` in favor of a PSR-3 logger.

I've also deprecated the public `AbstractExecutor::log()` method because I believe that the executor should not act as a logger.

Note that logging was not covered by tests at all, so the large part of this PR was actually adding tests. 🙈 

See https://github.com/doctrine/data-fixtures/pull/455#discussion_r1426888383 for a discussion.

### DIFF
--- a/src/Executor/AbstractExecutor.php
+++ b/src/Executor/AbstractExecutor.php
@@ -4,15 +4,21 @@ declare(strict_types=1);
 
 namespace Doctrine\Common\DataFixtures\Executor;
 
+use Closure;
 use Doctrine\Common\DataFixtures\FixtureInterface;
 use Doctrine\Common\DataFixtures\OrderedFixtureInterface;
 use Doctrine\Common\DataFixtures\Purger\PurgerInterface;
 use Doctrine\Common\DataFixtures\ReferenceRepository;
 use Doctrine\Common\DataFixtures\SharedFixtureInterface;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\Persistence\ObjectManager;
 use Exception;
+use Psr\Log\AbstractLogger;
+use Psr\Log\LoggerInterface;
+use TypeError;
 
-use function get_class;
+use function get_debug_type;
+use function is_callable;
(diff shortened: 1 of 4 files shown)


src/Executor/AbstractExecutor.php
```diff
@@ -78,22 +84,105 @@ public function getPurger()
...
+                public function log($level, $message, array $context = []): void
+                {
+                    $this->logger->log($level, $message, $context);
+                }
+
+                public function __invoke(string $message): void
+                {
+                    Deprecation::trigger(
+                        'doctrine/data-fixtures',
+                        'TODO',
```
greg0ire: :point_up: :smile:

derrabus: fixed

derrabus:
cc @GromNaN

tests/Common/DataFixtures/Executor/AbstractExecutorTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $executor = $this->bootstrapExecutor();
+        $logs     = [];
+
+        $this->expectDeprecationWithIdentifier('https://github.com/doctrine/data-fixtures/pull/462');
+
+        $executor->setLogger(static function (string $message) use (&$logs): void {
+            $logs[] = $message;
+        });
+
+        $executor->log('Something happend.');
```
greg0ire: `s/happend/happened` here and below

GromNaN (approved):
LGTM.

The command [`doctrine:fixtures:load`](https://github.com/doctrine/DoctrineFixturesBundle/blob/3.5.x/Command/LoadDataFixturesDoctrineCommand.php) will need an update to inject a `Symfony\Component\Console\Logger\ConsoleLogger` instead. The output will become more verbose.

greg0ire (approved):

PR #464, opened 2023-12-19
Make the class parameter mandatory for references

VincentLanglet:


greg0ire (approved):

tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
```diff
@@ -96,7 +98,7 @@ public function testReferenceReconstruction(): void
 
         $roleFixture->load($em);
         // first test against managed state
-        $ref = $referenceRepository->getReference('admin-role');
+        $ref = $referenceRepository->getReference('admin-role', Role::class);
```
derrabus, line 101: Should we have made these adjustments to the tests on 1.x already?

derrabus:
Please remember to update the UPGRADE.md file.

tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
VincentLanglet: Yes and no, depends on the test strategy.
The "old way" to call getReference must be tested on 1.x too.

On 1.x test should have been duplicated with
- The old way, working but triggering deprecation, and a `@group legacy`
- The new way working without deprecation.

Seems like no test were added/duplicated in https://github.com/doctrine/data-fixtures/pull/409, so only the old way were tested. Now on 2.x tests are updated.

SenseException (approved):

tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
derrabus: So, we're missing coverage for the non-deprecated way in 1.x currently? We should fix that first, I suppose.

VincentLanglet: I did https://github.com/doctrine/data-fixtures/pull/465

VincentLanglet:
Friendly ping @derrabus @greg0ire ; can it be merged ?

derrabus (approved):

ThomasLandauer:
This change is not documented at https://symfony.com/bundles/DoctrineFixturesBundle/current/index.html

VincentLanglet:
> This change is not documented at [symfony.com/bundles/DoctrineFixturesBundle/current/index.html](https://symfony.com/bundles/DoctrineFixturesBundle/current/index.html)

DoctrineFixtureBundle only support data-fixture v2 since few hours.
PR need to be done to update https://github.com/doctrine/DoctrineFixturesBundle/blob/3.7.x/docs/index.rst

derrabus:
Please open an issue (or a PR) on the bundle's repository.

PR #475, opened 2024-05-05
Setup dependabot

greg0ire:


greg0ire:
@derrabus, may I asked what your intent was with https://github.com/doctrine/data-fixtures/pull/453 ? I believe it's causing [the build failure](https://github.com/doctrine/data-fixtures/actions/runs/8958431245/job/24602645510?pr=475#step:7:28)

derrabus:
I wanted to test against upcoming releases (PHP 8.3, ORM 3, DBAL 4).

greg0ire:
I believe this is caused by https://github.com/sebastianbergmann/phpunit/commit/63183f242af98a6fe17700a60062ebc49e0f5325

What should we do?
- Use the workaround?
- Revert #453?
- Tweak config so that the "dev" build only allows dev versions of PHP, and `doctrine/*`?

derrabus:
> Use the workaround?

Let's not use mocks here. My proposal: #476

> Tweak config so that the "dev" build only allows dev versions of PHP, and `doctrine/*`?

In general: I think it's fine if the dev job fails from time to time because of changes upstream.

But we could of course also pin PHPUnit like we do on other repos. Maybe have dependabot bump our tools then?

greg0ire:
> But we could of course also pin PHPUnit like we do on other repos. Maybe have dependabot bump our tools then?

Sounds like a good idea 👍

derrabus (approved):

Issue #329, opened 2019-11-25
TopologicalSorter still breaks purge with foreign key constraints


videni:
I have this issue today  at `doctrine/data-fixtures    v1.3.2 `, what I did wrong? 

Order.orm.xml
```
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xmlns:gedmo="http://gediminasm.org/schemas/orm/doctrine-extensions-mapping"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                                      http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">

    <entity name="Acme\Bundle\OrderBundle\Entity\Order" table="acme_order">
        <id name="id" column="id" type="integer">
            <generator strategy="AUTO" />
        </id>
        <one-to-one field="review" target-entity="Acme\Bundle\ReviewBundle\Entity\OrderReview" mapped-by="reviewSubject"/>
        <one-to-one field="report" target-entity="Acme\Bundle\ReportBundle\Entity\Report" mapped-by="order"/>
    </entity>
</doctrine-mapping>
```

OrderReview.orm.xmln
```
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping">
    <entity name="Acme\Bundle\ReviewBundle\Entity\OrderReview" table="acme_order_review">
     <id name="id" column="id" type="integer">
            <generator strategy="AUTO" />
        </id>
        <one-to-one field="reviewSubject" inversed-by="review" target-entity="Acme\Bundle\OrderBundle\Entity\Order"/>
    </entity>
</doctrine-mapping>

```

Report.orm.xml
```
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:gedmo="http://gediminasm.org/schemas/orm/doctrine-extensions-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                                      http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">

    <entity name="Acme\Bundle\ReportBundle\Entity\Report" table="acme_auto_report"  >
        <id name="id" column="id" type="integer">
[...]

PR #481, opened 2024-06-20
Fix order of sorting for cyclic dependencies

esserj:
Fix order of sorting for cyclic dependencies where multiple dependencies are assumed to always
go after the cyclic one

In the test case:
Lets assume none of the FK allow cascading deletes

Removing Node 1 would be blocked by Node 2
Removing Node 2 would be blocked by Node 3 and Node 5
Removing Node 3 would be blocked by Node 4 -> cyclic through 4 -> 2 -> 3
Removing Node 4 would be blocked by Node 2 -> cyclic through 2 -> 3 -> 4
Removing node 5 is not blocked by any nodes

So We need to first remove Node 5, which is not blocked by any node
And Node 1 is not part is a cyclic dependency itself, but Node 2 references it, so Node 1 must always come somewhere after Node 2, which is part of a cyclic dep so Node 1 should then come after the other nodes

The rest of the nodes [2, 3, 4] are locked in cyclic dep. so we cannot really resolve them in a favourable way, unless we
try to identify which dependencies are blocking and which ones are not (deleting/nulling FKs or not), only listing the blocking dependencies should also resolve the issue, making the chance of success higher, this is however done by the Purger

### DIFF
--- a/src/Sorter/TopologicalSorter.php
+++ b/src/Sorter/TopologicalSorter.php
@@ -186,7 +186,7 @@ EXCEPTION
         $this->sortedNodeList[] = $definition->value;
     }
 
-    public function getDefinition(string $dependency, Vertex $definition): Vertex
+    private function getDefinition(string $dependency, Vertex $definition): Vertex
     {
         if (! isset($this->nodeList[$dependency])) {
             throw new RuntimeException(sprintf(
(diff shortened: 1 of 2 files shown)


esserj:
This will fix https://github.com/doctrine/data-fixtures/issues/329

esserj:
It might also fix https://github.com/doctrine/data-fixtures/pull/127
Because truncate would no longer be done on a table that is still referenced elsewhere

mvorisek:
Is this addressed in https://github.com/doctrine/orm/blob/2.20.5/src/Internal/TopologicalSort.php ?

esserj:
Nope, It is not really an issue there for us, but I guess it could be of the same entities are used for removal in a single UOW commit

PR #506, opened 2024-12-02
Relax type of identifiers

greg0ire:
ORM and ODM and other things having an implementation of ReferenceRepository do not necessarily agree on what an identifier should be.
data-fixtures should not really care either so long as the identifier can be used with serialize().
Let us relax the type of identifers to mixed.

### DIFF
--- a/src/ReferenceRepository.php
+++ b/src/ReferenceRepository.php
@@ -73,7 +73,7 @@ class ReferenceRepository
      * @param object $reference Reference object
      * @param object $uow       Unit of work
      *
-     * @return array
+     * @return mixed
      */
     protected function getIdentifier(object $reference, object $uow)
     {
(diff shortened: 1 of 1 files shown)


SenseException:
But doesn't `mixed` include non-serializable values too? Or do I confuse something with the types?

greg0ire:
It does, but I don't think there is a `serializable` type in PHP, is there? There is a `Serializable` interface, but here, we can deal with an array for instance.

Since [`serialize()` accepts `mixed`](https://www.php.net/manual/en/function.serialize.php), I guess `mixed` is the best we can do here.

SenseException (approved):
> Since [serialize() accepts mixed](https://www.php.net/manual/en/function.serialize.php), I guess mixed is the best we can do here.

This is probably the same relict as php functions returning false in case of an error and was transferred to newer PHP versions for compatibility.

PR #514, opened 2024-12-16
Remove fixed issue from baseline

greg0ire:
The commit that fixed this issue passed the build because it was created in parallel to 33b858a.


SenseException (approved):
Looks like this may also fix failing a build in #513 :+1:

PR #516, opened 2025-01-13
Fix error when use numeric for reference name

### LINKED ISSUE #512: Using numerics any reference name causes error with v1.8.1
### Bug Report

|    Q                                        |   A
|-------------------------------------------- | ------
| Version                                     | 1.8.1

#### Summary

After upgrade from 1.7.0 to 1.8.1 reference name that is numeric does not work, even when it is typed string.

#### Current behavior

Using numeric reference name causes error with 1.8.1, but not with 1.7.0

#### Expected behavior

Using numeric reference name working for both 1.7.0 and 1.8.1

#### How to reproduce

With 1.8.1
```php
$this->setReference('a12345', $anyEntity); // works
$this->setReference('12345', $anyEntity); // throws error 
```

```
 Doctrine\Common\DataFixtures\ReferenceRepository::setReferenceIdentity(): A  
  rgument #1 ($name) must be of type string, int given, called in /srv/vendor  
  /doctrine/data-fixtures/src/Event/Listener/ORMReferenceListener.php on line  
   52                       
```
With 1.7.0
```php
$this->setReference('a12345', $anyEntity); // works
$this->setReference('12345', $anyEntity); // works too
```

ovitores:
Fix #512 

The method `ReferenceRepository::setReferenceIdentity(string $name, mixed $identity, string $class): void`  typed the parameter `$name` with `string` and numeric reference names is cast to integers when are recovered from `ReferenceRepository::getReferenceNames($object)` method in previous lines, so when pass the value of numeric name in integer with a parameter typed with string then fails.

So I propose a fix to cast names to string in `ReferenceRepository::getReferenceNames($object)` when are returned

### DIFF
--- a/src/ReferenceRepository.php
+++ b/src/ReferenceRepository.php
@@ -12,6 +12,7 @@ use OutOfBoundsException;
 
 use function array_key_exists;
 use function array_keys;
+use function array_map;
 use function sprintf;
 
 /**
--- a/src/ReferenceRepository.php
+++ b/src/ReferenceRepository.php
@@ -192,7 +192,7 @@ class ReferenceRepository
             return [];
         }
 
-        return array_keys($this->referencesByClass[$class], $reference, true);
+        return array_map('strval', array_keys($this->referencesByClass[$class], $reference, true));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


derrabus:
LGTM, but we should fix this on 1.8.x.

ovitores:
> LGTM, but we should fix this on 1.8.x.

Ok, create bew PR in that branch... #517

derrabus:
Closing in favor of #517 then.

PR #517, opened 2025-01-13
Fix error when use numeric for reference name

ovitores:
Fix https://github.com/doctrine/data-fixtures/issues/512

The method ReferenceRepository::setReferenceIdentity(string $name, mixed $identity, string $class): void typed the parameter $name with string and numeric reference names is cast to integers when are recovered from ReferenceRepository::getReferenceNames($object) method in previous lines, so when pass the value of numeric name in integer with a parameter typed with string then fails.

So I propose a fix to cast names to string in ReferenceRepository::getReferenceNames($object) when are returned

### DIFF
--- a/src/ReferenceRepository.php
+++ b/src/ReferenceRepository.php
@@ -13,6 +13,7 @@ use OutOfBoundsException;
 
 use function array_key_exists;
 use function array_keys;
+use function array_map;
 use function get_class;
 use function sprintf;
 
@@ -35,7 +36,7 @@ class ReferenceRepository
      * List of named references to the fixture objects
      * gathered during fixure loading
      *
-     * @phpstan-var array<class-string, array<string, object>>
+     * @phpstan-var array<class-string, array<string|int, object>>
      */
     private array $referencesByClass = [];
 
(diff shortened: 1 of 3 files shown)


src/ReferenceRepository.php
```diff
@@ -35,7 +36,7 @@ class ReferenceRepository
      * List of named references to the fixture objects
      * gathered during fixure loading
      *
-     * @phpstan-var array<class-string, array<string, object>>
+     * @phpstan-var array<class-string, array<string|int, object>>
```
greg0ire, line 39: I have updated this to reflect the reality you described, which lead me to more static analysis fixes.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @ovitores !

pepeh:
Hi @greg0ire @SenseException 

I’m running into the exact same problem on 2.0.2. I noticed the fix got back-ported in 1.8.x, but it doesn’t look like it’s been applied to the 2.x branch yet. Could you clarify if there’s a plan to merge this change into 2.0, or if there’s another workaround for 2.0.2 in the meantime?

Thanks!

greg0ire:
Yeah, we usually do this after releasing (1.8.2 is not released yet)

PR #518, opened 2025-01-21
Allow doctrine persistence v4

mrVrAlex:
Since doctrine orm v3.3 package support both version of persistence then make sense allow it here.


derrabus (approved):
I've ran the tests locally and they pass with Persistence 4. However, our CI won't pick up Persistence 4 unless doctrine/mongodb-odm#2696 has been resolved.

PR #524, opened 2025-03-21
Run tests on PHP 8.4

greg0ire:


derrabus (approved):

derrabus:
is the test failure related?

SenseException:
```
There was 1 PHPUnit test runner warning:

1) No cache directory configured, result of static analysis for code coverage will not be cached
```

Tests weren't failing but I also get the warning locally. It must be related to the dev stability in the workflow matrix. I don't get this warning in phpunit when I install only stable packages with composer.

greg0ire:
This is why I created #525

PR #528, opened 2025-03-22
Use PHPStan 2

greg0ire:


composer.json
```diff
@@ -32,7 +32,7 @@
         "doctrine/mongodb-odm": "^1.3.0 || ^2.0.0",
         "doctrine/orm": "^2.14 || ^3",
         "fig/log-test": "^1",
-        "phpstan/phpstan": "1.12.21",
+        "phpstan/phpstan": "1.12.21 || 2.1.8",
```
derrabus:
```suggestion
        "phpstan/phpstan": "2.1.8",
```

🤔

derrabus (approved):

composer.json
greg0ire: Oh I had no idea 7.4 was supported, that's great!

PR #544, opened 2025-07-07
Add purge mode to MongoDB Purger

GromNaN:
Currently, the MongoDB purger drops the collection in order to purge it. Then it uses the SchemaManager to recreate its indexes.

But this approach isn't complete as other elements need to be created:
- The search indexes (requires ODM v2.8+ with https://github.com/doctrine/mongodb-odm/pull/2630)
- The auto encryption metadata (coming in https://github.com/doctrine/mongodb-odm/pull/2779)

Adding a new strategy that uses `deleteMany` to keep the collection instead of dropping and recreating everything. This is very important for Auto Encryption as the encryption keys are generated randomly each time an encrypted collection is created.

 For the ORM, the purge mode in set by a command option [`purge-with-truncate`](https://github.com/doctrine/DoctrineFixturesBundle/blob/901d0e84d05f49c94765e95782b277ca14393133/src/Command/LoadDataFixturesDoctrineCommand.php#L118). We have to add such option in the [`doctrine:mongodb:fixtures:load` command](https://github.com/doctrine/DoctrineMongoDBBundle/blob/5.4.x/src/Command/LoadDataFixturesDoctrineODMCommand.php)

[PHPORM-364](https://jira.mongodb.org/browse/PHPORM-364)

### DIFF
--- a/src/Purger/MongoDBPurger.php
+++ b/src/Purger/MongoDBPurger.php
@@ -5,25 +5,15 @@ declare(strict_types=1);
 namespace Doctrine\Common\DataFixtures\Purger;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
-use LogicException;
+
+use function method_exists;
 
 /**
  * Class responsible for purging databases of data before reloading data fixtures.
  */
 final class MongoDBPurger implements MongoDBPurgerInterface
 {
-    /**
-     * Purge the collections using deleteMany(). Don't create them.
-     */
-    public const PURGE_MODE_DELETE = 1;
-
-    /**
-     * Drop the collections when purging, then recreate them.
-     */
-    public const PURGE_MODE_DROP = 2;
-
-    /** @var self::PURGE_MODE_* $mode */
-    private int $purgeMode = self::PURGE_MODE_DELETE;
+    private MongoDBPurgeMode $purgeMode = MongoDBPurgeMode::Drop;
 
     /**
      * Construct new purger instance.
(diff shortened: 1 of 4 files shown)


src/Purger/MongoDBPurger.php
```diff
@@ -5,12 +5,26 @@
...
+     * Purge the collections using deleteMany(). Don't create them.
+     */
+    public const PURGE_MODE_DELETE = 1;
+
+    /**
+     * Drop the collections when purging, then recreate them.
+     */
+    public const PURGE_MODE_DROP = 2;
+
+    /** @var self::PURGE_MODE_* $mode */
```
stof:
```suggestion
    /** @var self::PURGE_MODE_* */
```

src/Purger/MongoDBPurger.php
```diff
@@ -5,12 +5,26 @@
...
 {
+    /**
+     * Purge the collections using deleteMany(). Don't create them.
+     */
+    public const PURGE_MODE_DELETE = 1;
+
+    /**
+     * Drop the collections when purging, then recreate them.
+     */
+    public const PURGE_MODE_DROP = 2;
```
stof: should this be an enum instead of integer constants in this class ? This is exactly the use case for enums. (ORMPurger uses class constants because enums were not available yet)

GromNaN: Thanks for the suggestion. The package requires PHP 8.1 so we can use enums. The type is improved that way.

stof (approved):

stof (approved):

SenseException (approved):

src/Purger/MongoDBPurger.php
```diff
@@ -38,15 +55,41 @@ public function getObjectManager(): DocumentManager
 
     public function purge(): void
     {
-        $metadatas = $this->dm->getMetadataFactory()->getAllMetadata();
-        foreach ($metadatas as $metadata) {
+        match ($this->purgeMode) {
+            MongoDBPurgeMode::Delete => $this->purgeWithDelete(),
+            MongoDBPurgeMode::Drop => $this->purgeWithDrop(),
+        };
```
GromNaN, line 61: I realize that I shouldn't have implemented the new mode like this. I should simply have created a new class that implements `MongoDBPurgerInterface` with the different behavior.

PR #552, opened 2025-08-13
Add DryRunORMExecutor class for dry-run support

Maxcastel:
This PR adds the DryRunORMExecutor class, allowing fixtures to be executed in a dry-run mode.

This class is intended to be used by bundles like DoctrineFixturesBundle to provide a --dry-run option.

**Related PR**:

- [DoctrineFixturesBundle PR #512](https://github.com/doctrine/DoctrineFixturesBundle/pull/512)

### DIFF
--- a/src/Executor/DryRunORMExecutor.php
+++ b/src/Executor/DryRunORMExecutor.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Common\DataFixtures\Executor;
+
+/**
+ * This executor allows to execute (and indirectly, print) SQL statements without
+ * actually committing them to the database
+ */
+final class DryRunORMExecutor extends AbstractExecutor
+{
+    use ORMExecutorCommon;
+
+    /** @inheritDoc */
+    public function execute(array $fixtures, bool $append = false): void
+    {
+        $executor = $this;
+        $this->em->beginTransaction();
+        try {
+            if ($append === false) {
+                $executor->purge();
+            }
+
+            foreach ($fixtures as $fixture) {
+                $executor->load($this->em, $fixture);
+            }
+
+            $this->em->flush();
+        } finally {
+            $this->em->rollBack();
+            $this->em->close();
+        }
+    }
+}
--- a/tests/Common/DataFixtures/Executor/DryRunORMExecutorTest.php
+++ b/tests/Common/DataFixtures/Executor/DryRunORMExecutorTest.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Common\DataFixtures\Executor;
+
+use Doctrine\Common\DataFixtures\Executor\DryRunORMExecutor;
+use Doctrine\Common\DataFixtures\FixtureInterface;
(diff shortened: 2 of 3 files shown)


greg0ire (commented):
Please write docs.

Also, I'm a bit surprised that there is such an amount of mocking for testing something that interacts with a database, but I cannot blame you for copying `ORMExecutorTest`.

src/Executor/DryRunORMExecutor.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Common\DataFixtures\Executor;
+
+/**
+ * Class responsible for executing data fixtures in dry run mode.
```
greg0ire: Let's not paraphrase the class' name and do some explaining instead

```suggestion
 * This executor allows to execute (and indirectly, print) SQL statements without
 * actually committing them to the database
```

tests/Common/DataFixtures/Executor/DryRunORMExecutorTest.php
```diff
@@ -0,0 +1,102 @@
...
+use Doctrine\Common\DataFixtures\Executor\DryRunORMExecutor;
+use Doctrine\Common\DataFixtures\FixtureInterface;
+use Doctrine\Common\DataFixtures\Purger\ORMPurgerInterface;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Tests\Common\DataFixtures\BaseTestCase;
+use PHPUnit\Framework\MockObject\MockObject;
+
+/**
+ * Test DryRun Fixture executor.
```
greg0ire: Let's remove this? 🤣

src/Executor/DryRunORMExecutor.php
Maxcastel: Done

tests/Common/DataFixtures/Executor/DryRunORMExecutorTest.php
Maxcastel: Done

Maxcastel:
@greg0ire
I’ve applied the requested changes and fixed the jobs.

greg0ire:
> Please write docs.

Please add documentation inside the docs directory

greg0ire (approved):

greg0ire:
I think this got closed because 2.0.x got destroyed.

greg0ire:
Note, this should be rebased on 2.2.x once it exists: https://github.com/doctrine/data-fixtures/pull/552

Maxcastel:
PHPUnit - PHP 8.1 still failing. Trying to fix it. 

I have the same version locally as in CI (PHP 8.1.33 - PHPUnit 10.5.45), but I don't get any errors locally.

greg0ire:
In your last attempt, the failing build is tagged with `lowest`, so maybe try `composer update --prefer-lowest` if you want to reproduce it locally.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @Maxcastel !

PR #574, opened 2025-10-12
Maintenance

greg0ire:


composer.json
GromNaN, line 17: Doctrine ORM and MongoDB ODM require PHP 8.1, we can increase the required version here also and require PHPUnit 10+

greg0ire: 1.8.x is soon to be unmaintained, I don't think we should bump this requirement, especially not in a patch version.

tests/Common/DataFixtures/ProxyReferenceRepositoryTest.php
```diff
@@ -33,12 +40,15 @@ public static function setUpBeforeClass(): void
...
 
+    #[IgnoreDeprecations]
     public function testReferenceEntry(): void
     {
         $em   = $this->getMockSqliteEntityManager();
         $role = new TestEntity\Role();
         $role->setName('admin');
         $meta = $em->getClassMetadata(self::TEST_ENTITY_ROLE);
+
+        // getPropertyAccessor() is not available with Doctrine < 3.4
```
derrabus:
```suggestion
        // getPropertyAccessor() is not available with ORM < 3.4
```

SenseException (approved):

PR #585, opened 2025-11-08
Use Native Lazy Objects with ODM when possible

greg0ire:
Not doing so is deprecated.
Note that without this patch, the CI is red.

### DIFF
--- a/tests/Common/DataFixtures/Purger/MongoDBPurgerTest.php
+++ b/tests/Common/DataFixtures/Purger/MongoDBPurgerTest.php
@@ -18,6 +18,8 @@ use function class_exists;
 use function dirname;
 use function method_exists;
 
+use const PHP_VERSION_ID;
+
 class MongoDBPurgerTest extends BaseTestCase
 {
     public const TEST_DOCUMENT_ROLE = Role::class;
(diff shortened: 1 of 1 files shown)


greg0ire:
@GromNaN at first I try putting the setting of proxies in the `else` clause of the `if` I was introducing, and was met with

```
TypeError: ProxyManager\Configuration::setProxiesTargetDir(): Argument #1 ($proxiesTargetDir) must be of type string, null given, called in /home/runner/work/data-fixtures/data-fixtures/vendor/doctrine/mongodb-odm/src/Configuration.php on line 657
```

I did not reproduce that locally. Here is the job: https://github.com/doctrine/data-fixtures/actions/runs/19191579564/job/54866771487#step:7:23

I find it weird that you still have to specify proxy-related stuff when enabling lazy objects.

GromNaN:
Indeed, there is a bug that I fix in https://github.com/doctrine/mongodb-odm/pull/2884

greg0ire:
Great, thank you! I think it would also make sense to deprecate calling methods that configure proxies when native lazy objects are enabled in ODM 2.15.x

SenseException (approved):

PR #586, opened 2025-11-08
phpcr-odm tests

dbu:
* add necessary dependencies to run the phpcr-odm tests in github actions
* fix phpcrpurger to implement interface and to use DocumentManagerInterface instead of DocumentManager.

### DIFF
--- a/src/Purger/PHPCRPurgerInterface.php
+++ b/src/Purger/PHPCRPurgerInterface.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace Doctrine\Common\DataFixtures\Purger;
 
+use Doctrine\ODM\PHPCR\DocumentManagerInterface;
+
 interface PHPCRPurgerInterface extends PurgerInterface
 {
     /**
      * Set the DocumentManager instance this purger instance should use.
      */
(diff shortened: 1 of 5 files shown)


src/Purger/PHPCRPurgerInterface.php
```diff
@@ -4,10 +4,12 @@
...
 
+use Doctrine\ODM\PHPCR\DocumentManagerInterface;
+
 interface PHPCRPurgerInterface extends PurgerInterface
 {
     /**
      * Set the DocumentManager instance this purger instance should use.
      */
-    public function setDocumentManager(DocumentManager $dm): void;
+    public function setDocumentManager(DocumentManagerInterface $dm): void;
```
dbu, line 14: this would be a BC break if it would have been possible to use the interface, but due to what the phpcrpurger previously implemented, it was not possible to use the interface i would say.

greg0ire: This causes an issue reported by PHPStan: if `PHPCRPurgerInterface` accepts the interface, then all implementations should accept it as well if not more.

greg0ire (approved):

SenseException (approved):
Less errors caught in the baseline 👍

